PR #2, opened 2021-10-27
New Token Transactions API

dievazqu:
## Description
Defining the new interface for fetching token transactions from the blockchain-api


src/schema.ts
```diff
@@ -206,6 +215,73 @@ export const typeDefs = gql`
...
+  """
+  A modified copy of the models of above. Except the new one support multiple tokens (using address instead of code to identify the currency)
+  """
+  type TokenAmount {
+    value: Decimal!
+    tokenAddress: Address!
+    localAmount: LocalMoneyAmount
+  }
+
+  enum TokenTransactionTypeV2 {
```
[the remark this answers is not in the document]
dievazqu: It wasn't used. But Gonza said that we should replace `ESCROW` with `INVITE`, so I will do that.

src/schema.ts
```diff
@@ -206,6 +215,79 @@ export const typeDefs = gql`
...
+    comment: String
+  }
+
+  """
+  TODO: Add more fields once we understand what useful information we can serve
+  """
+  interface TokenExchangeMetadata {
+    title: String
+    subtitle: String
+    exchangeName: String
```
gnardini, line 254: I'm not sure about this string, we might end up changing it for an enum or something but I guess it's fine for now :)

gnardini (approved):
LGTM! :)

src/schema.ts
```diff
@@ -206,6 +215,73 @@ export const typeDefs = gql`
     cursor: String!
   }
 
+  """
```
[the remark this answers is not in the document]
silasbw: it's just a string, right?  do you need to do anything special besides reading a file?

src/apolloServer.ts
```diff
@@ -2,7 +2,10 @@ import { ApolloServer } from 'apollo-server-express'
 import { BlockscoutAPI } from './blockscout'
 import CurrencyConversionAPI from './currencyConversion/CurrencyConversionAPI'
 import { logger } from './logger'
-import { resolvers, typeDefs } from './schema'
+import { resolvers } from './resolvers'
+import { readFileSync } from 'fs';
+
+const typeDefs = readFileSync('src/schema.graphql', 'utf-8');
```
silasbw, line 8: node.js includes os-independent path tools.  e.g., see [`path.join`](https://nodejs.org/api/path.html#pathjoinpaths)

silasbw: ...i think you want:

```ts
path.resolve(__dirname, 'schema.graphql')
```

src/schema.graphql
```diff
@@ -0,0 +1,218 @@
...
+  firstCursor: String
+  lastCursor: String
+}
+
+type Query {
+  tokenTransactionsV2(
+    address: Address!
+    tokens: [Address]
+    localCurrencyCode: String
+  ): TokenTransactionsV2
```
jeanregisser, line 199: Do we want pagination?
We actually never implemented in v1 but the structure was there to avoid a breaking change.

src/schema.graphql
```diff
@@ -0,0 +1,218 @@
...
+  comment: String
+}
+
+"""
+TODO: Add more fields once we understand what useful information we can serve
+"""
+interface TokenExchangeMetadata {
+  title: String
+  subtitle: String
+  exchangeName: String
```
jeanregisser, line 150: What's `exchangeName`?

jeanregisser: Do we need `image` here too?

src/schema.graphql
```diff
@@ -0,0 +1,218 @@
...
+  fees: [FeeV2]
+  metadata: TokenTransferMetadata
+}
+
+type TokenExchangeV2 implements TokenTransactionV2 {
+  type: TokenTransactionTypeV2!
+  timestamp: Timestamp!
+  block: String!
+  inAmount: TokenAmount!
+  outAmount: TokenAmount!
```
jeanregisser, line 177: This will work for simple exchanges but will limit us when more than 2 tokens are transferred in a transaction
I think we could merge `TokenTransferV2` and `TokenExchangeV2` and introduce a an array property containing the transfers and their direction (incoming or outgoing).
What do you think?

src/schema.graphql
dievazqu: I think we are ok with the current model. In the future when adding pagination, if we simply add new fields there won't be breaking changes. So, we could add a new field under `TokenTransactionsV2` and it should work.

src/schema.graphql
dievazqu: @gnardini suggested that we could include the name of the DEX used, that's what `exchangeName` is for.
I can add `image` :)

src/schema.graphql
dievazqu: I'm not sure about merging those types, but that might be because I don't understand the possible use cases.
Maybe we could add the new type that you are describing once we have the current use case? It shouldn't be hard to add a new type with its new model.
In the mean time, if you prefer I can change all the `TokenAmount` for `TokenAmount[]` so we can handle more that one token per transfer.

src/schema.graphql
gnardini: Maybe `exchangeName` is not the right field. What I thought is that we might want to show in the UI if the exchange was made using Ubeswap or Sushiswap for example, but maybe we won't need it or we will want this to look differently. I'm fine removing for now if we're worried about this field

jeanregisser: I see, maybe `appId`?

src/schema.graphql
jeanregisser: Yes and no, there are different ways to achieve pagination with GraphQL: https://graphql.org/learn/pagination/
For instance cursor pagination requires a specific structure which is hard to introduce without a breaking change. It's what the v1 implementation was proposing we do.
But I guess we don't really need it. 👍

src/schema.graphql
dievazqu: I'm gonna remove it until we have a clear view of what we want to do here.

src/schema.graphql
jeanregisser: I guess that's fine for now, I wanted to go a bit further in this new API, but let's do things incrementally. 👍

jeanregisser (approved):

src/schema.graphql
dievazqu: I see, I don't understand exactly how the pagination would be implemented with that structure. Is there some built-in pagination in GraphQL? I don't think so.
So I think we can add a `pageInfo` field under `TokenTransactionsV2` model to implement the same thing. Maybe I'm missing something 😅

PR #8, opened 2021-11-01
Deploy to alfajores

silasbw:


.github/workflows/workflow.yaml
```diff
@@ -21,14 +21,17 @@ jobs:
       - run: yarn
       - run: yarn build
       - run: yarn test
-  lint:
```
gnardini, line 24: Just curious, why did you delete the lint job?

gnardini (approved):
Left a Q regarding linting :)

.github/workflows/workflow.yaml
jh2oman: Looks like `yarn test` also runs eslint https://github.com/valora-inc/blockchain-api/blob/main/package.json#L9

jh2oman (approved):

.github/workflows/workflow.yaml
silasbw: I should have pulled that into a separate PR.  `yarn test` includes linting (and the `actions/setup-node@v2` action automatically parses the errors and annotates the PR).

PR #28, opened 2021-11-03
Creating DB for blockchain-api with historical prices table

dievazqu:


app.alfajores.yaml
```diff
@@ -12,3 +12,4 @@ env_variables:
   WEB3_PROVIDER_URL: 'https://alfajores-forno.celo-testnet.org'
   # Comma separated list of Secrets to load.
   SECRET_NAMES: 'projects/1067724576910/secrets/blockchain-api/versions/latest'
+  DB_SECRET: 'projects/1067724576910/secrets/blockchain-api-db/versions/latest'
```
silasbw, line 15: `SECRET_NAMES` can be a comma separated list of secrets to load, so you don't need to add a bunch of new env vars and the handle code for them.  So do:

```
SECRET_NAMES: 'projects/1067724576910/secrets/blockchain-api/versions/latest,projects/1067724576910/secrets/blockchain-api-db/versions/latest'
```

see https://github.com/valora-inc/blockchain-api/blob/main/src/index.ts#L19

src/index.ts
```diff
@@ -15,6 +16,18 @@ const PORT: number = Number(process.env.PORT) || 8080
 const INTERFACE: string = process.env.INTERFACE || '0.0.0.0'
 
 async function main() {
+  if (process.env.DB_SECRET) {
```
silasbw, line 19: see my comment above.  you'd move this below the secret loading code and do something like:

```
if (process.env.DB_HOST) {
```

I'd also suggest prepending `DB_` to the values in the secret (think of the secret values as a .env file stored and managed securely).

silasbw (changes requested):
Can you update the README.md w/ instructions for running postgres in a docker container?

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '2'
```
silasbw, line 1: do you need a compose file to manage a single container?  fine to add, but I'm curious about the benefits.

package.json
```diff
@@ -63,7 +65,7 @@
     "prettier": "^2.4.1",
     "ts-jest": "^27.0.7",
     "tsc-watch": "^4.2.3",
-    "typescript": "^3.8.3"
+    "typescript": "^4.2.4"
```
silasbw, line 68: go to `^4.4.4`?  or is there a reason you stopped here?

gnardini (commented):
Small detail + lets delete the `package-lock.json`, we use `yarn` here :)

app.alfajores.yaml
dievazqu: I'm not sure if handling all the secrets by adding them as environment variables is the best approach.
If by error we use the same variable name (and since we don't have all the secrets names visible, this could happen without detecting it), it will override one of them.
Since each secret will contain its own model, I think handling them separately is even more clear to understand what is expected to be obtained from each one. That's why I didn't use the current loading mechanism.

docker-compose.yml
dievazqu: In the future, we could add any dependency in this docker compose. So by simply running `docker-compose up`  we will have all the dependencies set up.

silasbw: I'd remove this until it's actually required.

app.alfajores.yaml
silasbw: I'm not sure I understand your concerns, but I'd like to follow an established pattern because it works well elsewhere.

We can discuss changing the pattern, but I want to change it everywhere if we do.

dievazqu: Ok, I will use current pattern.
I will give you an example of my concern. Let's say in the future we want to add a new db. If we are using: `DB_HOST`, `DB_USER`, etc as env var names, the new db secrets will have to know that and use different names to avoid overriding them. So, in order to create a new db, you need to have context of the current secrets, and the secrets aren't visible, so there is no easy way to do this.
My suggestion is to handle each secret independently, so we avoid this issue.

docker-compose.yml
dievazqu: I think it's already helpful and it's not a requirement for running blockchain-api locally. What is a requirement is to have a db to connect with. Actually that's the requirement that docker helps us solve.

silasbw: Sounds good.  Note, my feedback wasn't about `docker` it was about `docker-compose` specifically.

app.alfajores.yaml
silasbw: That makes sense, but that exists w/o this specific secret injection paradigm when using `env` files and non-namespaced env var names.  I think, however, that's what most people who use dotenv style configuration would expect and why we choose to follow that paradigm.

dievazqu: Yep, that's true it's not specific to secret injection. Using environment variables in any case have this concern. But using it for secrets have the downside that they aren't fully visible, so there are more chances for bugs.

silasbw (approved):

PR #33, opened 2021-11-04
Fix Apollo Server

gnardini:
Fix an issue with the newest version of `apollo-server-express`. Just a matter of calling `await apolloServer.start()`


silasbw (approved):
:hot_pepper: :ship:

PR #34, opened 2021-11-04
Add setup docs to README

gnardini:


.env.example
```diff
@@ -0,0 +1,9 @@
+DEPLOY_ENV=local
+FIREBASE_PROJECT_ID=celo-mobile-alfajores
```
silasbw, line 2: `<put real value>`

.env.example
```diff
@@ -0,0 +1,9 @@
...
+EXCHANGE_RATES_API=https://apilayer.net/api
+BLOCKSCOUT_API=https://alfajores-blockscout.celo-testnet.org/graphql
+FIREBASE_DB=https://celo-org-mobile-alfajores.firebaseio.com
```
silasbw, line 5: `<put real value>`

README.md
```diff
@@ -8,6 +8,26 @@ Install dependencies:
 yarn
 ```
 
+Copy the example .env file and check if there are any values you want to complete. If you need to fill in any secrets they are located in GCP's `Secret Manager` under `blockchain-api`.
```
silasbw, line 11: I'd remove the Secret Manager part.

README.md
```diff
@@ -8,6 +8,26 @@ Install dependencies:
...
+```
+cp .env.example .env
+```
+
+You also need a service account for the Firebase connection. You can create one from the GCP IAM console following these steps (ignore all except the last one if you already have one):
+
+- Open GCP's `celo-mobile-alfajores` project
```
silasbw, line 18: I'd remove the Valora specific details (like project name)

package.json
```diff
@@ -3,9 +3,6 @@
   "description": "A cloud service that stores and serves data about blockchain activity",
   "version": "0.0.1",
   "private": true,
-  "engines": {
```
silasbw, line 6: I added this to get an obvious and early failure when dependabot tries to bump dependencies that require newer versions of node.  Can we leave it in?  Feel free to bump it to a newer version if we can actually run on newer node.

silasbw (approved):
A w/ comment about the engines declaration

package.json
dievazqu: I understand that this field is advisory only, so I'm ok with removing it.

dievazqu (approved):
LGTM

package.json
gnardini: I had removed it because I thought it was not working with it, but it is now so I clearly did something else wrong yesterday 🤔  I re-added it in the meantime but it would be nice to update it eventually :)

PR #40, opened 2021-11-08
Storing token prices in DB

dievazqu:


src/cron.ts
```diff
@@ -1,6 +1,29 @@
...
-  // This function must return in 10 minutes.
-  logger.info('Should be updating prices...')
+import {
+  createNewManager,
+  getConfigForEnv,
+} from '@valora/exchanges'
+
+export async function updatePrices(db: Knex) {
+  logger.debug("Updating prices")
+  const config = getConfigForEnv(process.env.DEPLOY_ENV ?? 'local')
```
silasbw, line 11: I think it's clearer to export the specific configs from `@valora/exchanges` and require that value specifically instead of trying to map `DEPLOY_ENV` (which probably has specific meaning across different projects) to specific config values.  E.g.,

```ts
const config = exchanges.configs[process.env.EXCHANGES_CONFIG]
```

dievazqu: Yeah, this was to test locally, I'm gonna create a new environment variable for that instead of using `DEPLOY_ENV`.
I can export each config instead of `getConfigForEnv` if you think it's clearer.

dievazqu: Sorry, this PR is not ready to be reviewed yet. I'll ping #eng-pr-reviews once it's ready. :)

app.mainnet.yaml
```diff
@@ -13,3 +13,4 @@ env_variables:
   WEB3_PROVIDER_URL: 'https://forno.celo.org'
   # Comma separated list of Secrets to load.
   SECRET_NAMES: 'projects/1027349420744/secrets/blockchain-api/versions/latest,projects/1027349420744/secrets/blockchain-api-db/versions/latest'
+  EXCHANGES_ENV: 'mainnet'
```
gnardini, line 16: Is there any case where `EXCHANGES_ENV` would be different than `DEPLOY_ENV`? Just wondering if we need two separate flags

src/prices/PriceUpdater.ts
```diff
@@ -0,0 +1,28 @@
+import { Knex } from 'knex'
+import { logger } from '../logger'
+
+import { createNewManager, getConfigForEnv } from '@valora/exchanges'
+
+export async function updatePrices(db: Knex) {
```
gnardini, line 6: This looks good but will only track prices from here on out. I think we should find a solution that will give us historic prices as well in order to be able to show prices in the past.

Most contracts have an event they emit with their reserves. Uniswap pairs  https://github.com/Ubeswap/ubeswap/blob/master/contracts/uniswapv2/UniswapV2Pair.sol have a `Sync` event which has `reserve0` and `reserve1` fields. If we index these events in the indexer we can then use that to create the price graph the way it looked at any point in time. We can then run a script that calculates token prices since inception every one hour/day or so.

So:
1) What do you think of this approach?
2) Do you think we should merge this to get the transaction feed working in the short term and then work on adding backwards compatibility afterwards? It's only backend work so we should be able to do it without app changes.

app.mainnet.yaml
dievazqu: `DEPLOY_ENV` can have different values while running locally (`local`), in tests or in the e2e (`e2e`).
And actually we only support `alfajores` or `mainnet` for the exchage library, that's why I decided to use a different env var.  :)

src/prices/PriceUpdater.ts
dievazqu: I like that idea! But I think we should merge this first and then work on the past prices.
Also, I'm not sure about updating prices directly to the blockchain db from the indexer.
How do you feel about exposing an endpoint in the indexer that calculates prices for a given date, and then consuming that endpoint from the blockchain-api?

gnardini: Sounds good on merging this first.

On exposing an endpoint I'm unsure. If we do that, the `indexer` will have extra variable traffic, which it doesn't have now. Right now the indexer is a single instance that reads events and stores them in a db, so the instance usage is pretty stable. If we have a peak of users the indexer is unaffected. If we add an endpoint then we will need to consider how to scale the indexer and we'd also be adding additional complexity around services interacting with each other. This is going the microservices route which brings it's own set of complexities with it.

I understand not wanting to have different services writing to the same instance, but in that case how would you feel about giving the blockchain-api access to a read-only replica of the indexer db? Sounds like that might simplify the work?

PR #42, opened 2021-11-08
Add endpoint to fetch user balances

gnardini:
Added an endpoint that will be used in this PR: https://github.com/valora-inc/wallet/pull/1386 to fetch the token balances for users.


silasbw (commented):
probably a dumb question: do you know if blockscout is likely to support all the tokens Valora will?  do we care?

gnardini:
Blockscout returns all tokens that implement the ERC20 interface, it does no filtering. It just indexes every token that exists and returns their balance, even if they are not listed on any exchange.

silasbw (approved):
:artist:

dievazqu (approved):
LGTM

PR #47, opened 2021-11-11
Moving transactions to legacy folder

dievazqu:


silasbw (approved):
We will eventually remove the legacy code?

PR #49, opened 2021-11-16
Adding PricesService

dievazqu:


src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+    const prevPrice = new BigNumber(prevPriceRow.price)
+    const nextTimestamp = new Date(nextPriceRow.at).getTime()
+    const nextPrice = new BigNumber(nextPriceRow.price)
+
+    if (nextTimestamp === prevTimestamp) {
+      return prevPrice
+    }
+
+    // Linnear estimation
+    return prevPrice
```
gnardini, line 97: I think this should be `nextPrice`? We want to do:
```
(nextT - queryT) * nextP + (queryT - prevT) * prevP
----------------------------------------------------
                     nextT - prevT
```

If I understood correctly I think nextPrice and prevPrice are swapped in the current implementation.

The test looks good and is passing though so I might be reading this wrong 🤔

test/prices/PricesService.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    mockGetExchangeRate.mockReturnValue(1)
+  })
+
+  afterEach(async () => {
+    await db.destroy()
+    process.env = OLD_ENV
+  })
+
+  it('should return expected price', async () => {
+    await db(tableName).insert({
```
gnardini, line 48: two optional small suggestions: 
- Insert rows with other tokens to make sure they aren't considered.
- For these things I like to make them config based, so like creating an array of objects which have token, price and dateOffset and then doing
```
for (const rowInfo of rowsToInsert) {
  await db(tableName).insert({
     base_token: mockcUSDAddress,
      token: rowInfo.token,
      price: rowInfo.price,
      at: new Date(mockDate + rowInfo.dateOffset).toISOString(),
  })
}
```

I think this helps readability a bit when understanding the test as looking at the `rowsToInsert` array we can see only the fields that actually change without boilerplate around it.

test/prices/PricesService.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  let priceService: PricesService
+  const OLD_ENV = process.env
+
+  beforeEach(async () => {
+    jest.clearAllMocks()
+    db = await initDatabase()
+    process.env.EXCHANGES_ENV = 'test'
+    // @ts-ignore
+    priceService = new PricesService(db, mockExchangeAPI)
+    mockGetExchangeRate.mockReturnValue(1)
```
gnardini, line 39: Maybe add one test with a different exchange rate?

gnardini (commented):
Great job! Just left a few nits here and there :)

src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const TABLE_NAME = 'historical_token_prices'
+
+export default class PricesService {
+  db: Knex
+  exchangeAPI: ExchangeRateAPI
+  cUSDAddress: string
+
+  constructor(db: Knex, exchangeAPI: ExchangeRateAPI) {
+    this.db = db
```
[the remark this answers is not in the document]
dievazqu: Will try :)

src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+      throw new Error(
+        `Couldn't find entries in the db to calculate cUSD prices for ${token} at ${date}`,
+      )
+    }
+
+    return this.estimatePrice(prevPriceRow, nextPriceRow, date)
+  }
+
+  // It returns a linnear estimation of the price using previous and next known prices.
+  private estimatePrice(prevPriceRow: any, nextPriceRow: any, date: Date) {
```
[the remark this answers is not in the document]
dievazqu: Will do

src/prices/PricesService.ts
dievazqu: Actually the formula should be:
```
(nextT - queryT) * prevP + (queryT - prevT) * nextP
----------------------------------------------------
                     nextT - prevT
```

Think about it like this:
```
nextT - queryT
 ```
^^ tells you how close the queryT is to the nextT, but inversely (when queryT is equal to nextT the value is 0, not 1)
So we use 1 minus that as the weight, which is actually the inverse weight of the other price, that's why they are swapped.

Comment (dievazqu) [reply src/prices/PricesService.ts:97]: We need to find a function `price(t)` that
`price(prevT) = prevPrice`
`price(nextT) = nextPrice`
`price(t) = alpha(t) * prevPrice + (1 - alpha(t)) * nextPrice (with alpha(t) between 0 and 1)` <-- this is a weight sum

Then `alpha(t)` must be that:
`alpha(prevT) = 1` so `price(prevT) = prevPrice`
`alpha(nextT) = 0` so `prices(nextT) = nextPrice`

So `alpha(t) = (nextT - t) / (nextT - prevT)` <-- this is the linear estimation

Finally:
````
price(t) = alpha(t) * prevPrice + (1 - alpha(t)) * nextPrice
price(t) = (nextT - t) / (nextT - prevT) * prevPrice + (1 - ((nextT - t) / (nextT - prevT)) * nextPrice
price(t) = (nextT - t) * prevPrice / (nextT - prevT) + ((nextT - prevT) - (nextT - t)) * nextPrice / (nextT - prevT)
price(t) = ((nextT - t) * prevPrice  + (t - prevT) * nextPrice) / (nextT - prevT)
```

Hope this clarifies instead of confusing even more 😅

src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+    return await this.exchangeAPI.getExchangeRate({
+      sourceCurrencyCode: this.cUSDAddress,
+      currencyCode: localCurrency,
+      timestamp: date.getTime(),
+    })
+  }
+}
+
+function getcUSDFromConfig(): string {
+  if (!process.env.EXCHANGES_ENV) {
```
silasbw, line 116: this doesn't _seem_ like the _best_ place to handle (essentially) CLI arguments.  I think index.ts is probably a better place to assert that process.env.EXCHANGES_ENV is present and to get the cUSD token address.

src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+import { logger } from '../logger'
+
+const TABLE_NAME = 'historical_token_prices'
+
+export default class PricesService {
+  db: Knex
+  exchangeAPI: ExchangeRateAPI
+  cUSDAddress: string
+
+  constructor(db: Knex, exchangeAPI: ExchangeRateAPI) {
```
silasbw, line 14: see comment below, but I'd suggest passing in the cUSDAddress from where this is constructed

src/prices/PricesService.ts
```diff
@@ -0,0 +1,129 @@
...
+  ): Promise<BigNumber> {
+    return await this.exchangeAPI.getExchangeRate({
+      sourceCurrencyCode: this.cUSDAddress,
+      currencyCode: localCurrency,
+      timestamp: date.getTime(),
+    })
+  }
+}
+
+function getcUSDFromConfig(): string {
```
silasbw, line 115: Consider removing this function: see comment below and above.

test/prices/PricesUpdater.test.ts
```diff
@@ -39,6 +39,7 @@ describe('PricesUpdater#updatePrices', () => {
     dateNowSpy = jest.spyOn(Date, 'now').mockImplementation(() => mockDate)
 
     db = await initDatabase()
+    process.env.EXCHANGES_ENV = 'test'
```
silasbw, line 42: remove if you like my comments above

test/prices/PricesService.test.ts
dievazqu: I have changed the way it inserts rows in the DB. I think it's clearer without the array, but let me know if you still thinks the array config approach is better.

silasbw (approved):

gnardini (approved):
Looks great :)

test/prices/PricesUpdater.test.ts
dievazqu: Yep, this is still needed, this is for the `PricesUpdater` (in this PR I just moved that line, I didn't add it). But now that we have yargs, I will change that in another PR, so the environment parsing is done in the `index.ts`

test/prices/PricesService.test.ts
gnardini: Yup, all I wanted is to look at the inserts without the boilerplate, this does so very nicely :D

PR #53, opened 2021-11-18
Introduce some yargs

silasbw:
Given the proliferation of new arguments, it would be useful to bring some
argument handling conventions to the code. The rest of our projects use yargs
for this. This isn't a full conversion to yargs, but a start.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,6 @@
     "dotenv": "^10.0.0",
     "express": "^4.17.1",
     "express-prom-bundle": "^6.3.4",
-    "knex": "^0.95.14",
-    "pg": "^8.7.1",
     "firebase-admin": "^10.0.0",
     "graphql": "^15.7.2",
     "knex": "^0.95.11",
@@ -47,7 +45,8 @@
     "prom-client": "^14.0.1",
     "utf8": "^3.0.0",
     "web3": "1.6.1",
-    "web3-eth-abi": "1.6.1"
+    "web3-eth-abi": "1.6.1",
+    "yargs": "^17.2.1"
   },
   "devDependencies": {
     "@google-cloud/nodejs-repo-tools": "^3.3.0",
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,33 +31,32 @@ async function parseArgs() {
     .option('exchange-rates-api-access-key', {
       description: 'API key for exchange-rates-api',
       type: 'string',
-      demandOption: true
+      demandOption: true,
     })
     .option('blockchain-db-host', {
       description: 'Blockchain DB host',
       type: 'string',
-      demandOption: true
+      demandOption: true,
     })
     .option('blockchain-db-database', {
       description: 'Blockchain DB database',
       type: 'string',
-      demandOption: true
+      demandOption: true,
     })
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -25,18 +26,44 @@ async function main() {
     Object.assign(process.env, await loadSecret(secretName))
   }
 
-  if (!process.env.EXCHANGE_RATES_API_ACCESS_KEY) {
-    throw new Error('Missing required EXCHANGE_RATES_API_ACCESS_KEY')
-  }
+  const argv = yargs
+    .env('')
```
silasbw, line 30: Note: yargs parses environment variables too!  so you can either do `--exchange-rates-api-access-key=foo` or `EXCHANGE_RATES_API_ACCESS_KEY=foo`.

package.json
```diff
@@ -38,16 +38,15 @@
     "dotenv": "^10.0.0",
     "express": "^4.17.1",
     "express-prom-bundle": "^6.3.4",
-    "knex": "^0.95.14",
```
silasbw, line 41: these were listed twice in package.json

src/index.ts
```diff
@@ -25,18 +26,43 @@ async function main() {
...
+      demandOption: true,
+    })
+    .option('blockchain-db-pass', {
+      description: 'Blockchain DB pass',
+      type: 'string',
+      demandOption: true,
+    })
+    .epilogue(
+      'Always specify arguments as environment variables. Not all arguments are ' +
+        'supported as CLI ones yet.',
```
jeanregisser, line 58: nit: ```suggestion
      'Always specify arguments as environment variables. Not all arguments are supported as CLI ones yet.',
```

File: src/index.ts (hunk 1 above)
Comment (jeanregisser) [reply src/index.ts:30]: Interesting trick, so this throws an exception when missing?

**Review by jeanregisser — APPROVED** [review]

File: src/index.ts [hunk 4]
```diff
@@ -25,18 +26,43 @@ async function main() {
...
-    !process.env.BLOCKCHAIN_DB_DATABASE ||
-    !process.env.BLOCKCHAIN_DB_USER ||
-    !process.env.BLOCKCHAIN_DB_PASS
-  ) {
-    throw new Error("Blockchain database secrets couldn't be obtained")
-  }
+async function main() {
+  const args = await parseArgs()
 
   const db = await initDatabase()
```
dievazqu, line 67: To initialize the DB we are using `process.env.BLOCKCHAIN_DB_HOST  and others. Should we change that and get those values from `args`?

src/index.ts
silasbw: Yes

src/index.ts
silasbw: Yes we should -- those changes are more invasive so i was going to do them in self-contained PRs.  This one was simply focused on adding yargs.  The hope is that authors of other PRs could know to use yargs while I loop back around and address things like you point out.  Seem reasonable?

dievazqu (approved):
LGTM

src/index.ts
dievazqu: Sure!

PR #58, opened 2021-11-20
Convert database options and port option to yargs

silasbw:


src/database/db.ts
```diff
@@ -2,16 +2,6 @@ import { Knex, knex } from 'knex'
 import { logger } from '../logger'
 
 async function checkAndMigrate(db: Knex) {
-  try {
```
silasbw, line 5: I think this delay is unnecessary or at least not super robust to races.  It doesn't retry, so if there's a race with the DB starting up (_e.g._, as part of e2e tests), it will fail.  Removing it seems to avoid the race, but if `db.migrate.latest` races to, I can add retry-wait logic here.

gnardini (approved):

PR #59, opened 2021-11-21
Convert EXCHANGES_ENV to yargs (and rename EXCHANGES_CONFIG)

silasbw:
EXCHANGES_CONFIG matches the naming of `@valora/exchanges/config`


silasbw:
I'll rebase the merge base on this PR after https://github.com/valora-inc/blockchain-api/pull/58 merges.

src/index.ts
```diff
@@ -30,6 +34,12 @@ async function parseArgs() {
       type: 'number',
       default: 8080,
     })
+    .option('exchanges-config', {
```
gnardini, line 37: This is not really the config, this is the key of a network -> config mapping that is used by the `exchanges` package. I find calling the network we use `config` a bit confusing tbh.

src/prices/PriceUpdater.ts
```diff
@@ -1,28 +1,25 @@
...
+} from '@valora/exchanges'
+
+export async function updatePrices({
+  db,
+  exchangeRateManager,
+  exchangeRateConfig,
+}: {
+  db: Knex
+  exchangeRateManager: ExchangeRateManager
+  exchangeRateConfig: ExchangeRateConfig
```
gnardini, line 16: The manager after your other PR exports the `cUSDTokenAddress`, right? In that case I think we don't need to pass the `exchangeRateConfig` here, we can just fetch it from the manager.

src/prices/PriceUpdater.ts
```diff
@@ -1,28 +1,25 @@
...
 
-import { createNewManager, configs } from '@valora/exchanges'
-
-export async function updatePrices(db: Knex) {
+import {
+  ExchangeRateManager,
+  Config as ExchangeRateConfig,
+} from '@valora/exchanges'
+
+export async function updatePrices({
```
gnardini, line 9: Question here, why receive only one parameter with an object instead of multiple parameters? From the caller POV  we're doing `await updatePrices({ db, exchangeRateManager, exchangeRateConfig })` instead of ` await updatePrices(db, exchangeRateManager, exchangeRateConfig)` so it's pretty similar in practice, and here we need to double declare them in order to add the types which looks a bit off to me.

I understand that in some cases, especially when there are optional fields this makes caller's life much simpler and clear (passing `undefined` values is not clear at all) but I'm not sure if it's actually helping in this case.

gnardini (commented):
Left a couple small comments, but changes look good

src/index.ts
silasbw: that's what it's called in the `@valora/exchanges`: https://github.com/valora-inc/exchanges/blob/main/src/config.ts

Happy to do a rename across all the repos, but I first want to make everything consistent.

src/prices/PriceUpdater.ts
silasbw: that's right -- i'll merge the other PR and publish the module and then loop back around and change this!

src/prices/PriceUpdater.ts
silasbw: oh, i was following what i thought the convention was in other repos: favor keyword arguments vs positional.

If you're asking about why folks like that convention, i think it's generally simple and well-defined: each function takes a single argument; if you want to pass what are logically multiple arguments use keyword arguments (i.e., objects with destructuring); argument order should never matter (i.e., not everyone uses autocomplete / intellsense, whatever).

The point of this PR isn't to try and redefine (or define a poorly defined) convention, so happy to do whatever.

src/prices/PriceUpdater.ts
silasbw: updated!

dievazqu (approved):

src/index.ts
dievazqu: Maybe `exchanges-config-key` or `exchanges-config-env`?

silasbw: I can rename everything across the two repos.  I'd purpose that we try to make things consistent in this PR (and I _think_ `exchanges-config` is consistent) and a follow-on PR can rename everything if that's what we want to do.

gnardini: In the other repo the config is the object that has `fullNodeUrl`, `tokenAddresses` and `ubeswap` and then there's a `Record<string, Config>` that holds them. This variable is this `string` so calling it `exchange-config` I think is misleading since this is the key with which you can get the config, not the config itself. So I guess `exchanges-config-key` is more accurate but imo the clearest would be `exchanges-config-network`.

silasbw: +1 to "network".  I'll update.

PR #63, opened 2021-11-23
Improve error handling when fetching exchange rates fails

gnardini:


src/currencyConversion/ExchangeRateAPI.ts
```diff
@@ -33,19 +34,30 @@ export default class ExchangeRateAPI extends RESTDataSource {
...
+        currencyCode,
+        date,
+      )
 
-    return new BigNumber(fetchedRate)
+      return new BigNumber(fetchedRate)
+    } catch (error) {
+      logger.error({
+        type: 'ERROR_FETCHING_EXCHANGE_RATE',
+        error: (error as Error)?.message,
```
silasbw, line 54: this should be an instance of `Error` (the `formatError` in `apolloServer.ts` is wrong), so just pass the `error` (not `error.message`).  We'll get much more helpful debugging context in the logs.

FYI, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bunyan/index.d.ts#L151-L156

src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -54,9 +54,16 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
     }
 
     const rates = await Promise.all(ratesPromises)
+    const undefinedRateIndex = rates.findIndex((rate) => !rate)
```
silasbw, line 57: nit: since the index isn't actually used, I find these this alternative slightly quicker to read and understand:

```ts
if (rates.indexOf(undefined) >= 0) {
  return null
}
```

if we're really talking about "falseyRateExists" (or something like that), then:

```ts
const falseyRateExists = rates.findIndex((rate) => !rate) >= 0
if (falseyRateExists) {
  return null
}
```

src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -125,7 +132,7 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
     toCode: string,
     timestamp?: number,
     impliedExchangeRates?: MoneyAmount['impliedExchangeRates'],
-  ): BigNumber | Promise<BigNumber> {
+  ): BigNumber | Promise<BigNumber | undefined> {
```
silasbw, line 135: Why does `getSupportedExchangeRate` return `Promise<BigNumber | undefined>` while `getExchangeRate` returns `Promise<BigNumber | null>`?  Is there a compelling reason to have slightly different return types?

(i'm doing my best to ignore the ignore the `BigNumber |` present here)

src/currencyConversion/ExchangeRateAPI.ts
```diff
@@ -33,19 +34,30 @@ export default class ExchangeRateAPI extends RESTDataSource {
     sourceCurrencyCode,
     currencyCode,
     timestamp,
-  }: CurrencyConversionArgs): Promise<BigNumber> {
+  }: CurrencyConversionArgs): Promise<BigNumber | undefined> {
```
silasbw, line 37: Similar to above: it would be nice to make these return types consistent: either `Promise<BigNumber | undefined>` or `Promise<BigNumber | null>`.  I have a preference for `null`, but it probably doesn't matter much.

src/currencyConversion/ExchangeRateAPI.ts
```diff
@@ -33,19 +34,30 @@ export default class ExchangeRateAPI extends RESTDataSource {
...
+      const fetchedRate = await this.queryExchangeRate(
+        sourceCurrencyCode || USD,
+        currencyCode,
+        date,
+      )
 
-    return new BigNumber(fetchedRate)
+      return new BigNumber(fetchedRate)
+    } catch (error) {
+      logger.error({
```
silasbw, line 52: Can you add a comment here on why it's better to log the error, suppress it, and return `undefined` than let the error bubble up to the ApolloServer (and have the default logger log it)?

src/resolvers.ts
```diff
@@ -209,10 +212,13 @@ export const resolvers = {
...
         impliedExchangeRates: moneyAmount.impliedExchangeRates,
       })
+      if (!rate) {
+        return null
+      }
       return {
         value: new BigNumber(moneyAmount.value).multipliedBy(rate).toString(),
         currencyCode: localCurrencyCode || 'USD',
-        exchangeRate: rate.toString(),
+        exchangeRate: rate?.toString(),
```
silasbw, line 221: doesn't every object have toString

src/prices/PricesService.ts
```diff
@@ -115,7 +118,7 @@ export default class PricesService {
   private async cUSDToLocalCurrency(
     localCurrency: string,
     date: Date,
-  ): Promise<BigNumber> {
+  ): Promise<BigNumber | undefined> {
```
silasbw, line 121: similar comment on undefined vs null

src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -34,7 +34,7 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
     currencyCode,
     timestamp,
     impliedExchangeRates,
-  }: CurrencyConversionArgs): Promise<BigNumber> {
+  }: CurrencyConversionArgs): Promise<BigNumber | null> {
```
dievazqu, line 37: I think it's better to throw an exception instead of returning `null` or `undefined`.
So when using this service we don't have to consider both cases (exception or `null` value) as a not normal execution. When probably we would handle both cases as the same.

src/resolvers.ts
```diff
@@ -209,10 +212,13 @@ export const resolvers = {
         timestamp: moneyAmount.timestamp,
         impliedExchangeRates: moneyAmount.impliedExchangeRates,
       })
+      if (!rate) {
```
dievazqu, line 215: Same as above

src/resolvers.ts
```diff
@@ -165,6 +165,9 @@ export const resolvers = {
         // released and most users update.
         sourceCurrencyCode: args.sourceCurrencyCode ?? USD,
       })
+      if (!rate) {
```
dievazqu, line 168: Following my other comment, I think we should try/catch `dataSources.currencyConversionAPI.getExchangeRate` and return `null` in case of an error.

src/currencyConversion/CurrencyConversionAPI.ts
dievazqu: The same applies for other services.

src/prices/PricesService.ts
```diff
@@ -34,6 +34,9 @@ export default class PricesService {
         localCurrency,
         date,
       )
+      if (!cUSDToLocalCurrencyPrice) {
```
dievazqu, line 37: Following: Neither we would have to add this

dievazqu (commented):
Overall looks good!
However, I think we should treat this situation as an exception.
But I don't feel really strong in my opinion, so I'm open to discuss it.

gnardini:
@dievazqu @silasbw 
I changed to have the code throw the error again like Diego suggested and have the places that want to handle it catch it. I agree it's a better approach. This helped remove all the `undefined` and `null` issues too, sorry for the inconsistency there.

I did also add on the transfers endpoint a manual check for the `makerAmount` and `takerAmount` fields since those two were failing. If we can't fetch the exchange rate we just ignore those. This will help not return items we know will fail while not throwing 500 errors at the same time.

src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -29,6 +33,26 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
...
+  async getFromMoneyAmount({
+    moneyAmount,
+    localCurrencyCode,
+  }: {
+    moneyAmount: MoneyAmount
+    localCurrencyCode: string | undefined
+  }): Promise<LocalMoneyAmount> {
+    const rate = await this.getExchangeRate({
+      sourceCurrencyCode: moneyAmount.currencyCode,
+      currencyCode: localCurrencyCode || 'USD',
```
dievazqu, line 45: nit: extract `localCurrencyCode || 'USD'` to a variable to avoid the duplication

src/resolvers.ts
```diff
@@ -203,16 +216,21 @@ export const resolvers = {
...
-        currencyCode: localCurrencyCode || 'USD',
-        timestamp: moneyAmount.timestamp,
-        impliedExchangeRates: moneyAmount.impliedExchangeRates,
-      })
-      return {
-        value: new BigNumber(moneyAmount.value).multipliedBy(rate).toString(),
-        currencyCode: localCurrencyCode || 'USD',
-        exchangeRate: rate.toString(),
+      try {
+        const rate = await dataSources.currencyConversionAPI.getExchangeRate({
```
dievazqu, line 220: nit: we can call the new function `getFromMoneyAmount` here instead, right?

src/blockscout.ts
```diff
@@ -222,26 +226,53 @@ export class BlockscoutAPI extends RESTDataSource {
...
 
+    const filteredEvents = (
+      await Promise.all(
+        events.map(async (event) => {
+          if (!event.makerAmount && !event.takerAmount) {
+            return event
+          }
+
+          try {
+            await currencyConversionAPI.getFromMoneyAmount({
```
dievazqu, line 247: So basically, we are attempting the future calls here? And filter those transactions if any of them fails?
It doesn't seem to be the best approach. But I guess this will be changed with the new transaction feed implemention. So I'm ok with it.

src/prices/PricesService.ts
dievazqu: This is not needed anymore, right?

dievazqu (approved):
LGTM but I prefer someone else to take a look before merging.

src/blockscout.ts
```diff
@@ -222,26 +226,53 @@ export class BlockscoutAPI extends RESTDataSource {
...
       .filter((e) => e)
       .filter((event) => tokens.includes(event.amount.currencyCode))
       .sort((a, b) => b.timestamp - a.timestamp)
 
+    const filteredEvents = (
+      await Promise.all(
+        events.map(async (event) => {
+          if (!event.makerAmount && !event.takerAmount) {
+            return event
+          }
```
jeanregisser, line 244: Are we only checking exchanges?

src/blockscout.ts
```diff
@@ -222,26 +226,53 @@ export class BlockscoutAPI extends RESTDataSource {
...
             transaction: JSON.stringify(transaction),
-            error: (e as Error)?.message,
           })
         }
       })
       .filter((e) => e)
       .filter((event) => tokens.includes(event.amount.currencyCode))
       .sort((a, b) => b.timestamp - a.timestamp)
 
+    const filteredEvents = (
```
jeanregisser, line 239: We should add a comment explaining what we are doing here and why.

dievazqu:
> @dievazqu @silasbw I changed to have the code throw the error again like Diego suggested and have the places that want to handle it catch it. I agree it's a better approach. This helped remove all the `undefined` and `null` issues too, sorry for the inconsistency there.
> 
> I did also add on the transfers endpoint a manual check for the `makerAmount` and `takerAmount` fields since those two were failing. If we can't fetch the exchange rate we just ignore those. This will help not return items we know will fail while not throwing 500 errors at the same time.

I missed this message before reviewing the PR.
I'm not sure about the conversion fetching calls in the blockscout class. If the requests to apilayer are failing because the service has too much load, maybe those requests will pass (the ones in blockscout class) but then the real ones might fail (the ones that we do in the `MoneyAmount` resolver). So I think we could still miss some cases with this approach.
If the apilayer is failing consistenly for some specific addresses, this should work. But I don't know if we can assume that.

jeanregisser (commented):
Thanks!

src/blockscout.ts
gnardini: Yes, I think those are the one that were causing the app to crash, no?

gnardini:
@dievazqu The result from calling apilayer are cached, so when they are requested again later they will be just fetched from the cache, not queried from the API again, so we shouldn't have that issue. Does that sound good? Is there anything else that worries you about this?

dievazqu:
> @dievazqu The result from calling apilayer are cached, so when they are requested again later they will be just fetched from the cache, not queried from the API again, so we shouldn't have that issue. Does that sound good? Is there anything else that worries you about this?

Oh, I forgot the cache. Yep, that sounds great!

src/blockscout.ts
jeanregisser: Yep all good!

src/apolloServer.ts
```diff
@@ -28,7 +28,7 @@ export function initApolloServer({
       }
     },
     formatError: (error) => {
-      logger.error({
+      logger.error(error, {
         type: 'UNHANDLED_ERROR',
         error: error?.message,
```
silasbw, line 33: you can remove this (the logger will set `.msg` to `error.message`)

silasbw (approved):

PR #72, opened 2021-12-03
Transaction feed price service

dievazqu:
Calculating `TokenAmount.localAmount` using `PricesServices`


src/prices/PricesService.ts
```diff
@@ -115,12 +127,16 @@ export default class PricesService {
     )
   }
 
-  private async cUSDToLocalCurrency(
+  private async USDToLocalCurrency(
     localCurrency: string,
     date: Date,
   ): Promise<BigNumber> {
-    return await this.exchangeAPI.getExchangeRate({
-      sourceCurrencyCode: this.cUSDAddress,
```
gnardini, line 123: We used to pass an address but now a symbol? Looks like a weird change, does it not break anything?

gnardini (approved):
Looks good! Just a bit confused about the address vs symbol we pass as `sourceCurrencyCode`

src/prices/PricesService.ts
```diff
@@ -115,12 +127,16 @@ export default class PricesService {
     )
   }
 
-  private async cUSDToLocalCurrency(
+  private async USDToLocalCurrency(
```
jeanregisser, line 130: nit: functions should use camelCase
```suggestion
  private async usdToLocalCurrency(
```

jeanregisser (approved):
LGTM 👍

src/prices/PricesService.ts
dievazqu: I don't know why I did like that in the past 😅 
The latest version is the one that works :)

PR #83, opened 2021-12-13
Adding migration for cGLD prices

dievazqu:
Migrate historical cGLD prices from firebase to the blockchain-api database

### DIFF
--- a/src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
+++ b/src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
@@ -54,12 +54,18 @@ async function fetchHistoricalData(tokenAddresses: TokenAddresses) {
 async function fetchTokensAddresses(): Promise<TokenAddresses> {
   const snapshot = (await database.ref(`tokensInfo`).once('value')).val()
   const tokensInfoValue = Object.values(snapshot)
+  const cUSDAddress = getAddressForSymbol(tokensInfoValue, 'cUSD')
+  const celoAddress = getAddressForSymbol(tokensInfoValue, 'CELO')
+
+  if (!cUSDAddress || !celoAddress) {
+    throw new Error("Can't obtain token addresses");
+  }
   return {
-    cUSD: getAddressForSymbol(tokensInfoValue, 'cUSD'),
-    CELO: getAddressForSymbol(tokensInfoValue, 'CELO'),
+    cUSD: cUSDAddress,
+    CELO: celoAddress,
   }
 }
 
-function getAddressForSymbol(tokensInfo: any[], symbol: string): string {
+function getAddressForSymbol(tokensInfo: any[], symbol: string): string | undefined {
   return tokensInfo.find((token) => token.symbol === symbol)?.address
 }
--- a/src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
+++ b/src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
@@ -1,0 +1,56 @@
+
+import { Knex } from 'knex'
+import { database } from '../../firebase'
+
+interface TokenAddresses {
+  cUSD: string,
+  cEUR: string,
+  CELO: string
+}
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('historical_token_prices', (table) => {
+    table.string('fetched_from')
+  })
+
+  const tokenAddresses = await fetchTokensAddresses()
+  console.log(JSON.stringify(tokenAddresses))
+  const tuplesToInsert = await fetchHistoricalData(tokenAddresses)
+  await knex.batchInsert('historical_token_prices', tuplesToInsert)
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex('historical_token_prices')
+    .delete()
+    .where({fetched_from: '20211213154345_migration'})
+
+  await knex.schema.alterTable('historical_token_prices', (table) => {
+      table.dropColumn('fetched_from')
+  })
+}
+
+async function fetchHistoricalData(tokenAddresses: TokenAddresses) {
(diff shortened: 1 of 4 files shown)


src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
```diff
@@ -0,0 +1,65 @@
+import { Knex } from 'knex'
+import { logger } from '../../logger'
+import { database } from '../../firebase'
+
+interface TokenAddresses {
+  cUSD: string
+  CELO: string
+}
+
+// Once the migration is done in alfajores and mainnet, we could remove this file or at least silent its errors
```
gnardini, line 10: Why would this throw an error? Shouldn't even be run, right?

gnardini: Oh, or do you mean on e2e tests?

src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
```diff
@@ -0,0 +1,65 @@
...
+async function fetchTokensAddresses(): Promise<TokenAddresses> {
+  const snapshot = (await database.ref(`tokensInfo`).once('value')).val()
+  const tokensInfoValue = Object.values(snapshot)
+  return {
+    cUSD: getAddressForSymbol(tokensInfoValue, 'cUSD'),
+    CELO: getAddressForSymbol(tokensInfoValue, 'CELO'),
+  }
+}
+
+function getAddressForSymbol(tokensInfo: any[], symbol: string): string {
```
gnardini, line 63: This function technically returns `string | undefined`, no? I'm surprised Typescript didn't complain. Should never be undefined of course, just surprised

src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
```diff
@@ -0,0 +1,65 @@
...
+async function historicalPricesMigration(knex: Knex) {
+  const tokenAddresses = await fetchTokensAddresses()
+  const tuplesToInsert = await fetchHistoricalData(tokenAddresses)
+
+  await knex.batchInsert('historical_token_prices', tuplesToInsert)
+}
+
+async function fetchHistoricalData(tokenAddresses: TokenAddresses) {
+  const celoUSDSnapshot = Object.values(
+    (await database.ref(`exchangeRates/cGLD/cUSD`).once('value')).val(),
```
gnardini, line 42: Do you know how big this node is? Just wondering if it _could_ bring any issues. I guess we can deploy in alfajores and see 🤷‍♂️

gnardini, line 42: I think we need the `cGLD/cEUR` pair as well

gnardini (commented):
Looks good! I think we should just include the cEUR exchange rates as well

src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
dievazqu: Yep, I mean in e2e and unit tests.

src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
dievazqu: I tested it with alfajores and it worked. But I'm wondering if it works for mainnet 🤔 
Do you know if there is a way to stream a node in firebase?
It seems that we have around 25k entries in mainnet. It should work.

gnardini: Yeah, should be fine

dievazqu: Yep, the problem is that having `cGLD/cEUR` doesn't give us info about the USD price of the `cEUR` without any extra info.
Here are some ideas of how we should handle this:
1) We could fetch `cGLD/cEUR` and then for each entry we can calculate the `cGLD` USD price (using the data that we have in the db), and the multiply those values.
2) We could use the `Euro / Dollar` price: We could iterate for some dates, and use the exchange API to request the price at that moment. With this we could have more entries that we currently have in firebase, however, the stable tokens aren't 1:1 to the fiat currency, so this approach will be slightly inaccurate.

What do you think? Do you have a better idea for this?

dievazqu: Oh, another option would be to store the `cGLD/cEUR` price in the db, and change the way we calculate prices. Because right now we are only checking against USD price. But we could have some logic to build a path to get to USD.
I really don't like this approach, it increases the complexity of the calculation only to consider this case.

gnardini: Ah sorry I missed this. I don't really like the approach of changing the way we do prices either. I see you went the route of fetching the CELO/cUSD price which I think is a good one!

src/database/migrations/20211213154345_migrate_historical_prices_from_firebase.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  return await Promise.all(
+    Object.values(celoEurSnapshot).map(async (entry: any) => {
+      const isoDate = new Date(entry.timestamp).toISOString()
+      const priceRow = await knex<HistoricalPriceRow>(TABLE_NAME)
+        .where({
+          token: tokenAddresses.CELO,
+          base_token: tokenAddresses.cUSD,
+        })
+        .andWhere('at', '=', isoDate)
```
gnardini, line 62: Should this be `<=` instead? We might not have the exact date I think, right?

gnardini (commented):
Left a question, rest looks great!

gnardini (approved):
👏 👏

PR #93, opened 2022-01-05
Refactoring Transaction Feed classes

dievazqu:
Replacing `TransferCollection` and `TransfersNavigator` classes with a utility class (`TransferUtils`).
Moving methods from `InputDecoder` to `Input`.


gnardini (approved):
Looks good to me!

PR #104, opened 2022-01-26
Fix cUSD price behavior

dievazqu:
We are not storing historical prices for cUSD, since it's always 1.
However, we were querying the db for fetching cUSD price. Resulting in an error, and the localCurrency price wasn't being returned.


gnardini (approved):
Good catch!

PR #106, opened 2022-01-28
Improve logging and error handling

gnardini:
Changes:
- Pretty Print is not used in production anymore, just the json logger.
- Added retrying for Blockscout queries.
- Moved to a slightly larger instance class (we were in the smallest one and I saw a couple of memory-related errors).
- Revised the severity of logs.
- Made sure all logs were in json format.

I tested all this locally and deployed in alfajores

- Fixes [#1882](https://github.com/valora-inc/wallet/issues/1882)

### DIFF
--- a/src/blockscout.ts
+++ b/src/blockscout.ts
@@ -44,6 +44,7 @@ import {
   TokenSent,
 } from './events'
 import { TransactionAggregator } from './transaction/TransactionAggregator'
+import { Body } from 'apollo-datasource-rest/dist/RESTDataSource'
 export interface BlockscoutTransferTx {
   blockNumber: number
   transactionHash: string
@@ -117,7 +118,7 @@ export class BlockscoutAPI extends RESTDataSource {
           try {
             return await type.getEvent(transaction)
           } catch (e) {
-            logger.error({
+            logger.warn({
               type: 'ERROR_MAPPING_TO_EVENT_V2',
               transaction: JSON.stringify(transaction),
               error: (e as Error)?.message,
@@ -144,7 +145,7 @@ export class BlockscoutAPI extends RESTDataSource {
 
     await this.ensureContractAddresses()
 
-    const response = await this.post('', {
+    const response = await this.queryBlockscoutWithRetry({
       query: `
         query Transfers($address: AddressHash!) {
           # TXs related to cUSD or cGLD transfers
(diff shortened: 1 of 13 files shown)


src/helpers/Input.ts
```diff
@@ -31,8 +31,11 @@ export class Input {
 
     try {
       return coder.decodeParameters(abi, this.data)
-    } catch (e) {
-      logger.debug(`Error decoding input: ${(e as Error).message}`)
+    } catch (error) {
+      logger.warn({
+        type: 'INPUT_DECODE_ERROR',
+        error: (error as Error)?.message,
```
silasbw, line 37: I think you can include the error directly (vs `error.message`): https://github.com/trentm/node-bunyan#log-method-api

src/logger.ts
```diff
@@ -2,13 +2,14 @@ import * as Logger from 'bunyan'
...
 
-const consoleStream = new PrettyStream()
-consoleStream.pipe(process.stdout)
-
-const streams: Logger.Stream[] = [{ stream: consoleStream, level: 'info' }]
+const streams: Logger.Stream[] = []
 if (process.env['GAE_APPLICATION']) {
   const loggingBunyan = new LoggingBunyan()
   streams.push(loggingBunyan.stream('info'))
+} else {
```
silasbw, line 9: Great!  I've been annoyed at the pretty printing logs showing up in google cloud :pray:

src/utils.ts
```diff
@@ -145,3 +146,26 @@ export async function getContractKit(): Promise<ContractKit> {
     throw new Error('Failed to create contractKit instance')
   }
 }
+
+export async function runWithRetries<ReturnType>(
```
silasbw, line 150: should you move this to another PR?  i think it's worth more discussion and log improvements all look good.

src/blockscout.ts
```diff
@@ -198,12 +203,16 @@ export class BlockscoutAPI extends RESTDataSource {
     return transactions
   }
 
+  async queryBlockscoutWithRetry(body: Body) {
```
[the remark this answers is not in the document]
gnardini: It fails very often, so I figured it would be better to retry here already to avoid the round trips.

src/helpers/Input.ts
gnardini: Right, I figured out after changing this that is was the pretty logs that were causing the untidy logs, I thought it was this, I'll revert these changes

src/utils.ts
gnardini: Removed the retry from here. Blockscout queries fail _very_ often though so I figured we'd be doing the client a favor by retrying here, but I can get more data before coming back with this.

src/helpers/Input.ts
gnardini: Added the error serializer thing :)

src/resolvers.ts
```diff
@@ -219,7 +240,11 @@ export const resolvers = {
         })
         return { rate: rate.toNumber() }
       } catch (error) {
-        logger.error(error, { type: 'ERROR_CURRENCY_CONVERSION' })
+        logger.error({
+          error,
+          type: 'CURRENCY_CONVERSION_ERROR',
+          sourceCurrencyCode: args.sourceCurrencyCode,
```
dievazqu, line 246: I think it would be helpful to print all the args, specially the timestamp.

dievazqu: Also, maybe this is should be a `warn` instead of an `error`

gnardini: Hmm, I think it's an error, this is called directly from Valora and the app is quite limited without it since we can't show local amounts. At least it's something that we should track how often it fails.

gnardini: Good catch on logging all the args!

PR #118, opened 2022-03-04
feat: Update prices from blockchain-api

dievazqu:
## Description

This PR add 2 features:
* It updates current usd prices for every token in firebase every minute. (This replaces the cloud-function)
* It updates last day prices for every token in firebase every 10 minutes.


src/helpers/KnownAddressesCache.ts
```diff
@@ -16,8 +16,8 @@ class KnownAddressesCache {
   private knownAddresses: KnownAddresses = {}
 
   startListening(): void {
+    logger.info({ type: 'FETCHED_KNOWN_ADDRESSES' })
```
gnardini, line 19: 🤔  This doesn't look right?

src/helpers/TokenInfoCache.ts
```diff
@@ -21,8 +21,8 @@ class TokenInfoCache {
   private tokensInfo: TokensInfo = {}
 
   startListening(): void {
+    logger.info({ type: 'FETCHED_TOKENS_INFO' })
```
gnardini, line 24: Same as the other, if we want to move it here it needs to say `START_FETCHING_X` or something, if it uses the past tense it indicates that it loaded new values I think? At least that's what I understand from it

src/index.ts
```diff
@@ -94,16 +98,70 @@ async function main() {
...
+    db,
+    exchangeRateAPI,
+    exchangeRateManager.cUSDTokenAddress,
+  )
+
   const app = express()
 
   app.use(metricsMiddleware)
 
+  // What is this? lol
```
gnardini, line 118: I think this `robots.txt` file is to tell Google's indexing bots how to read/index your website. I'm not sure why we have it here though.

src/index.ts
```diff
@@ -94,16 +98,70 @@ async function main() {
...
   app.get('/robots.txt', (_req, res) => {
     res.type('text/plain')
     res.send('User-agent: *\nDisallow: /')
   })
 
-  app.get('/cron/update-prices', async (req, res) => {
+  app.get('/cron/update-current-prices', async (req, res) => {
+    // App Engine sets this header if and only if the request is from a cron.
```
gnardini, line 125: I don't think it's a big deal but should we validate the content of the header or something? Doesn't look like the check is doing much otherwise

src/prices/PricesUpdater.ts
```diff
@@ -2,34 +2,120 @@ import { Knex } from 'knex'
 import { logger } from '../logger'
 
 import { ExchangeRateManager } from '@valora/exchanges'
+import { updateFirebase } from '../firebase'
+import tokenInfoCache from '../helpers/TokenInfoCache'
+import PricesService from './PricesService'
 
-export async function updatePrices({
+const asyncPool = require('tiny-async-pool')
```
gnardini, line 9: Can we use the normal `import asyncPool from 'tiny-async-pool'` here?

src/prices/PricesUpdater.ts
```diff
@@ -2,34 +2,120 @@ import { Knex } from 'knex'
...
+
+const FIREBASE_NODE_KEY = '/tokensInfo'
+const MAX_CONCURRENCY = 30
+export const DAY = 1000 * 60 * 60 * 24
```
gnardini, line 13: nit: I think it would be clearer to name the const `ONE_DAY_IN_MS`

cron.yaml
```diff
@@ -1,5 +1,13 @@
 cron:
-  - description: 'Update prices'
-    url: /cron/update-prices
+  - description: 'Store current prices in the db'
+    url: /cron/store-prices
```
gnardini, line 3: For storing prices I think 10 minutes might be too much granularity? I imagine the db will fill up _quick_. How about doing it every one hour instead? And I also think maybe in a followup PR it would be nice to add a cron that deletes data older than 30 days and just leaves 1 record per day, no need to have so much old info imo.

Maybe I'm overestimating the amount of data and it's not so bad though, just thinking out loud.

src/index.ts
```diff
@@ -94,16 +98,70 @@ async function main() {
...
   app.use(metricsMiddleware)
 
+  // What is this? lol
   app.get('/robots.txt', (_req, res) => {
     res.type('text/plain')
     res.send('User-agent: *\nDisallow: /')
   })
 
-  app.get('/cron/update-prices', async (req, res) => {
+  app.get('/cron/update-current-prices', async (req, res) => {
```
gnardini, line 124: I see these three endpoints all follow a very similar approach and are a bit boilerplate-y. What do you think about:

- First, putting all the crons in their own file. I think we can create a sub-router in it and here do `app.use('/cron', cronRouter)`. See https://stackoverflow.com/a/55425436/16044077 for an example. This would clean up this `index` file quite a bit.
- Then, the only difference I see between the three endpoints is the function they call and the `type` they use for logging. Maybe we can create a function that receives a function builder and the type and does the rest?

gnardini (commented):
Great work! 😄

src/helpers/TokenInfoCache.ts
dievazqu: I wanted to avoid having this log every time a change was done in the node. I guess I can change the type. :)

src/index.ts
dievazqu: This was added by @silasbw the first time we did a cron. I think it's to prevent hitting this API directly. However, I also think this isn't adding much value. I'm ok with removing it, if @silasbw agrees.

src/prices/PricesUpdater.ts
dievazqu: I've tried, but I was having compiling issues. This was the only way I could make it work. 😕

gnardini: Oh, I think it's because we are missing the `@types/tiny-async-pool` dependency. See https://github.com/valora-inc/nft-viewer/blob/main/package.json#L62 for example

cron.yaml
dievazqu: Yep, I'm ok with doing this every hour.
Not sure about leaving only a record per day thing, this info is used for showing the estimated price in transactions. And I guess that we are showing transactions older than a month frequently.
However, we could create a cron to do that job, and then we tweak the numbers. Could we decide the numbers based on some metric? If we realize we are not showing transactions (or we show a really small amount of transactions) older than X time, we should use that value.
Doing some rough calculations, we would be storing less than 1Mb per day, I think we are ok with that amount at the moment, so there is no need to rush for that feature, but it's good to start thinking about it.

src/index.ts
dievazqu: Yep, I'll do that!

src/prices/PricesUpdater.ts
dievazqu: Oh! Yep, that worked! Thanks!

cron.yaml
gnardini: Yeah, that doesn't sound like a lot of info, it's probably simpler to just keep it to once an hour and revisit in the future if we need to :)

src/crons.ts
```diff
@@ -0,0 +1,71 @@
...
+
+function addEndpoint(
+  router: Router,
+  path: string,
+  errorType: string,
+  asyncFn: () => Promise<void>,
+) {
+  router.get(path, async (_, res) => {
+    try {
+      await asyncFn()
```
gnardini, line 20: I'm doubtful of its utility but I would leave the header check that Silas added for now, it doesn't hurt either I guess :)

gnardini (approved):
Amazing work! 🙌

PR #133, opened 2022-03-25
(feat): Adding pagination to Transaction Feed

dievazqu:
Now, when requesting for the transaction feed, the parameter `afterCursor` can be sent, which it's used to query blockscout to obtain the next page.


src/blockscout.ts
```diff
@@ -77,6 +89,46 @@ export interface BlockscoutTokenTransfer {
   value: string
 }
 
+const BLOCKSCOUT_QUERY = `
+query Transfers($address: AddressHash!, $afterCursor: String) {
+  # TXs related to cUSD or cGLD transfers
+  tokenTransferTxs(addressHash: $address, first: 100, after: $afterCursor) {
```
kathaypacific, line 95: i'm not sure if i understand correctly, does this hardcoded `first: 100` or the latter `first:10` in the `tokenTransfer` array mean that the number of results per page is fixed? was wondering if it'd be nice/possible for the consumer to decide this, or have this number in a config variable somewhere in blockchain-api to make it easier to find/update

src/blockscout.ts
```diff
@@ -133,55 +191,32 @@ export class BlockscoutAPI extends RESTDataSource {
...
-                  }
-                }
-              }
-            }
-          }
-        }
-      `,
-      variables: { address },
+      query: BLOCKSCOUT_QUERY,
+      variables: { address, afterCursor },
```
kathaypacific, line 215: will this be backwards compatible with older apps (people who haven't updated their app) if the older apps don't send an `afterCursor`? i *think* this looks fine from the JS side but i'm not sure what happens to the `BLOCKSCOUT_QUERY` string when it receives undefined for the `afterCursor`, just curious if you have already tried this scenario :)

test/blockscoutV2.test.ts
```diff
@@ -139,14 +139,30 @@ describe('Blockscout', () => {
...
-  it('should get all transactions and label them properly', async () => {
+  it('should get all transactions and label them properly when there is no afterCursor', async () => {
     const result = await blockscoutAPI.getTokenTransactionsV2(
       '0x0000000000000000000000000000000000007E57',
     )
 
     // Reversing for convenience to match the order in mock data
-    const transactions = result.reverse()
+    const transactions = result.transactions.reverse()
+    expect(transactions).toMatchSnapshot()
```
kathaypacific, line 149: as there is an existing snapshot for when the `afterCursor` is passed, what do you think about asserting that the outcome of `await blockscoutAPI.getTokenTransactionsV2(
      '0x0000000000000000000000000000000000007E57',
    )` deep equals `const result = await blockscoutAPI.getTokenTransactionsV2(
      '0x0000000000000000000000000000000000007E57',
      'afterCursor',
    )`?

src/blockscout.ts
kathaypacific: i think your tests cover this, but i'm not clear on the details of how we mock the blockscout endpoint so thought i'd check if you'd tried on a live env :)

kathaypacific (commented):
i just had a couple questions but looks really great!!

src/blockscout.ts
dievazqu: Yep, currently those are fix values. I can extract them to some variable so it's easier to change.
I'm not sure about letting the user set these values, because if these values are too big the query doesn't work, so I prefer to set these values to the current ones that we know they work.
Or, I guess we can let the user set any value lower than this, so the query should be faster...

src/blockscout.ts
dievazqu: Yep, I have tried it, and it works :)

test/blockscoutV2.test.ts
dievazqu: Sure :)

gnardini (approved):
Sorry for taking so long to review,  looks great!

PR #138, opened 2022-04-19
Adding peg prices mechanism

dievazqu:


src/helpers/TokenInfoCache.ts
```diff
@@ -41,6 +42,10 @@ class TokenInfoCache {
   getTokensAddresses() {
     return Object.keys(this.tokensInfo)
   }
+
+  getTokensInfo() {
+    return Object.assign([], this.tokensInfo)
```
gnardini, line 47: Hmm, why are we doing this? `this.tokensInfo` is an object but we're assigning it to an array?

gnardini (approved):
Small confusion and a nit, but looks great other than that 😄

src/helpers/TokenInfoCache.ts
dievazqu: My bad, it should be `Object.values(this.tokensInfo)`

PR #148, opened 2022-05-16
Add Sentry support

silasbw:


src/index.ts
```diff
@@ -132,6 +160,27 @@ async function main() {
   await apolloServer.start()
   apolloServer.applyMiddleware({ app, path: GRAPHQL_PATH })
 
+  if (sentryEnabled) {
+    // The Sentry error handler must be before any other error middleware and
+    // after all controllers
+    app.use(Sentry.Handlers.errorHandler())
+  }
+
+  app.use(
```
silasbw, line 169: ...i didn't see a pre-existing default error handler.

src/index.ts
```diff
@@ -80,6 +86,30 @@ async function parseArgs() {
...
+      integrations: [
+        // enable HTTP calls tracing
+        new Sentry.Integrations.Http({ tracing: true }),
+        // enable Express.js middleware tracing
+        new Tracing.Integrations.Express({ app }),
+      ],
+
+      // Set tracesSampleRate to 1.0 to capture 100%
+      // of transactions for performance monitoring.
+      // We recommend adjusting this value in production
```
cajubelt, line 106: should this be an environment variable too so we can set it lower for production?

cajubelt (approved):

src/index.ts
silasbw: Maybe?  I'm hoping we can just keep it at 100% (like we do for valora-rest-api).

PR #154, opened 2022-05-17
Upgrade alfajores to Node 16

silasbw:
Temporarily drop node engines semver while we run on Node 16 in alfajores and
Node 12 in mainnet.


jeanregisser (approved):
🚀

dievazqu (approved):

PR #156, opened 2022-05-18
Add Sentry tracesSampleRate CLI option and configure for 0.5

silasbw:


silasbw:
We're pushing too much to Sentry:

  https://sentry.io/organizations/valora-inc/stats/?dataCategory=transactions

dievazqu (approved):

PR #176, opened 2022-07-14
support NFT_transaction event

Samuel3Shin:
### Description
Adding NFT transaction event for NFT transaction query from wallet repo.
Currently, if a transaction contains at least one `ERC-721` token transfer, blockchain-API classifies that as a `NftReceived` or `NftSent` event. 

If the transaction has more `NFT received transfer` than  `NFT sent transfer` then it is `NftReceived` event.
Otherwise, it is  `NftSent` event.

If the numbesr of `NFT received transfer` and `NFT sent transfer` are the same, we decide the event according to the last transfer. If the last NFT transfer is `NFT received transfer`  then, it is  `NftReceived` event. 

Screen Shot 2022-07-21 at 2 09 57 PM

### Other changes
N/A

### Tested
With Mac Pro 2019

### How others should test
With local machine

### Related issues
- https://github.com/valora-inc/wallet/pull/2698
- https://linear.app/valora/issue/RET-329/support-nfts-in-the-home-feed

### Backwards compatibility
Yes


src/blockscout.ts
```diff
@@ -234,7 +240,10 @@ export class BlockscoutAPI extends RESTDataSource {
 
     const filteredUnknownTokens = transactions.filter((tx: Transaction) => {
       return tx.transfers.every((transfer: BlockscoutTokenTransfer) => {
-        return supportedTokens.has(transfer.tokenAddress.toLowerCase())
+        return (
+          supportedTokens.has(transfer.tokenAddress.toLowerCase()) ||
+          transfer.tokenType === 'ERC-721'
```
Samuel3Shin, line 245: It doesn't filter if the token is `ERC-721`. Do you guys think, it would cause some bug while classifying transactions???

dievazqu: I think this is ok. But maybe we could consider moving the filtering to the classifying step (like, leaving the transfer as an unknown transaction if uses an unsupported token.
But what you did seems to work :)

dievazqu (commented):
LGTM!
It would be nice if you could add a test for this event.
Maybe just adding a new ERC-721 transaction in the `mockTokenTxsV2.ts`.

src/blockscout.ts
Samuel3Shin: Yup, I think it would be better to consider that after this PR. Probably, when implementing `swap transaction event`??

Samuel3Shin: And I added the ERC-721 transaction in the `mockTokenTxsV2.ts` as you told me!

README.md
```diff
@@ -37,7 +37,7 @@ For the emulator:
 
 ```
 # Check if there are any values you need to complete
-cp emuluator.env .env
+cp emulator.env .env
```
kathaypacific, line 40: nice :D

src/events/NftReceived.ts
```diff
@@ -0,0 +1,38 @@
...
+          isNftReceived = true
+        }
+
+        if (transfer.fromAddressHash === this.context.userAddress) {
+          isNftReceived = false
+        }
+      }
+    }
+
+    return transaction.transfers.length >= 1 && isNftExist && isNftReceived
```
kathaypacific, line 25: could this be simplified to 
```
return transaction.transfers.some(transfer => transfer.tokenType === 'ERC-721' && transfer.toAddressHash === this.context.userAddress)
```
?

src/events/NftSent.ts
```diff
@@ -0,0 +1,38 @@
...
+          isNftSent = false
+        }
+
+        if (transfer.fromAddressHash === this.context.userAddress) {
+          isNftSent = true
+        }
+      }
+    }
+
+    return transaction.transfers.length >= 1 && isNftExist && isNftSent
```
kathaypacific, line 25: maybe the above simplification could also be made here?

kathaypacific (commented):
Looking great!!!

src/events/NftReceived.ts
Samuel3Shin: Some NFT transaction might have both `transfer.toAddressHash === this.context.userAddress` and `transfer.fromAddressHash === this.context.userAddress`,

in this situation, I want to decide that it is `Sent NFT` if I see `transfer.fromAddressHash === this.context.userAddress` in the last time in the for loop.

And I decide it is `Received NFT` if I see `transfer.toAddressHash === this.context.userAddress`  in the last time in the for loop.

If we use `transaction.transfers.some()`, all the transactions with both transfers might be classified as only one event

src/blockscout.ts
```diff
@@ -151,11 +155,15 @@ export class BlockscoutAPI extends RESTDataSource {
 
     const context = { userAddress }
 
+    // Order is important when classifying transactions.
+    // Think that below is like case statement.
```
Samuel3Shin, line 159: @kathaypacific 
This part is like `case statement`, so If we use `transaction.transfers.some()`, all the transactions with both transfers might be classified as only one event

src/events/NftSent.ts
Samuel3Shin: I replied to this in the above thread

src/events/NftReceived.ts
kathaypacific: ^ would you mind checking this? if it's true that there can be multiple ERC-721 tokens going to and from a user's address and we want to explicitly only let the last of these transfer decide the result of whether the NFT is sent or received, maybe it is nice to add a test for this. in the sample data you added in `mockTokenTxsV2.ts`, i see only 1 ERC-721 transfer in the list (which in my mind was what they all would look like but i am not sure now)

Samuel3Shin: `transaction.transfers.some(transfer => transfer.tokenType === 'ERC-721' && transfer.toAddressHash === this.context.userAddress)`

`transaction.transfers.reverse().some(transfer => transfer.tokenType === 'ERC-721' && transfer.toAddressHash === this.context.userAddress)`

I think the above two functions result in the same result whatsoever. This is because, even though it is reserved or not, it will return true if any element in the array passes the test. 

"The some() method tests whether at least one element in the array passes the test implemented by the provided function. It returns true if, in the array, it finds an element for which the provided function returns true; otherwise it returns false. It doesn't modify the array." (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

kathaypacific: oh you're totally right! i didn't think that ode through, what i wanted to do was a `reverse()` then loop through until the condition is hit then break the loop. so `some()` wasn't the right thing to use here, apologies! but thinking about it now again, the whole thing wasn't a very good suggestion anyway.

i had DM'ed you this but putting it here for visibility - if we are catering to the scenario of there being multiple nft's that can appear in the transfers, then letting the last transfer decide the result may not be the most correct. what do you think about logic like "a transaction is an NFT transaction if it contains NFT type transfers AND there is more NFT transfers that are to the user's address than from the user's address"?

.github/dependabot.yml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "npm"  # For dependabot npm is synonymous to yarn
-    directory: "/"
+  - package-ecosystem: 'npm' # For dependabot npm is synonymous to yarn
```
kathaypacific, line 3: are changes to this file intentional?

src/events/NftReceived.ts
```diff
@@ -12,18 +12,24 @@ export class NftReceived extends TransactionType {
     for (const transfer of transaction.transfers) {
       if (transfer.tokenType === 'ERC-721') {
         if (transfer.toAddressHash === this.context.userAddress) {
-          cntNftReceivedTransfer++;
+          cntNftReceivedTransfer++
```
kathaypacific, line 15: can we use `num` instead of `cnt`? i think this is more common in the projects. maybe `numNftsReceived`?

src/events/NftReceived.ts
```diff
@@ -12,18 +12,24 @@ export class NftReceived extends TransactionType {
...
       }
     }
 
-    return transaction.transfers.length >= 1 && (cntNftSentTransfer + cntNftReceivedTransfer > 0) && (cntNftSentTransfer == cntNftReceivedTransfer ? isNftReceived : cntNftSentTransfer < cntNftReceivedTransfer)
+    return (
+      transaction.transfers.length >= 1 &&
+      cntNftSentTransfer + cntNftReceivedTransfer > 0 &&
+      (cntNftSentTransfer == cntNftReceivedTransfer
+        ? isNftReceived
+        : cntNftSentTransfer < cntNftReceivedTransfer)
```
kathaypacific, line 31: could the return statement be simplified to `return numNftsReceived > numNftsSent`? these variables will be 0 otherwise, idk if the other conditions change the result of this condition?

kathaypacific (commented):
Almost there, looking great!! 🥳

kathaypacific (approved):
Approved pending my small comments from earlier :)

src/events/NftReceived.ts
Samuel3Shin: Yes, I changed it

src/events/NftReceived.ts
Samuel3Shin: If we change it to that we miss the transaction if `numNftsReceived` is 2, `numNftsSent` is 2. Is that what you intend!??

src/events/NftReceived.ts
Samuel3Shin: Yup, that's a very good idea. I changed the logic like your suggestion!

.github/dependabot.yml
Samuel3Shin: Yes, I ran prettier and it looks better

src/events/NftReceived.ts
Samuel3Shin: Yes, I changed it to `numNftsReceived` and `numNftsSent`!

src/events/NftReceived.ts
Samuel3Shin: Oh, actually, I can remove "cntNftSentTransfer + cntNftReceivedTransfer > 0".
But. I think we need to consider when "cntNftSentTransfer == cntNftReceivedTransfer" because there might be some cases when they are the same. Then we can classify it according to the last transfer.

kathaypacific: hum, i know you merged this already but i think that we should have a better rule than "let the last item decide". 

i'm thinking that we should view the transaction as a whole, and that order of transfers doesn't change the nature of the transaction - so in that sense, if we have a transaction with equal numbers of NFTs going in and out of an address, then that is one type of transaction (even if in one transaction, the last transfer is a "sent NFT" and in another transaction, the last transfer is "received NFT") - do you think this makes sense? for me it would be strange if sometimes we have NFT sent and sometimes we have NFT received in this case :/ 

we could implement this rule pretty easily if we want to, like to classify as NFT sent we require the net number of transfers with the user's address in the "from" field to be `>` 0, and NFT received we require the net number of transfers with the user's address in the "to" field to be `>=` 0.

PR #182, opened 2022-08-02
feat(NFT invite rewards): supporting NFT events

Samuel3Shin:
### Description
** added wallet client version check for backward compatibility. (https://github.com/valora-inc/blockchain-api/pull/182/commits/8db09e6678266e6a5a5f78bb10051926c0640f24)

Adding NFT transaction event for NFT transaction query from wallet repo.
Currently, if a transaction contains at least one `ERC-721` token transfer, blockchain-API classifies that as a `NftReceived` or `NftSent` event. 

If the transaction has more `NFT received transfer` than  `NFT sent transfer` then it is `NftReceived` event.
Otherwise, it is  `NftSent` event.

If the numbesr of `NFT received transfer` and `NFT sent transfer` are the same, we decide the event to be a `NftReceived` event. 

Screen Shot 2022-07-21 at 2 09 57 PM

### Other changes
N/A

### Tested
With Mac Pro 2019

### How others should test
With local machine

### Related issues
- https://github.com/valora-inc/wallet/pull/2751
- https://linear.app/valora/issue/RET-329/support-nfts-in-the-home-feed

### Backwards compatibility
Yes


src/blockscout.ts
```diff
@@ -149,13 +155,27 @@ export class BlockscoutAPI extends RESTDataSource {
       afterCursor,
     )
 
+    // For now, when you create a new transaction type other than TokenTransferV2, TokenExchangeV2
+    // You should do version check to take care of backward compatibility with wallet client.
+    const userInfo = await fetchFromFirebase(`registrations/${userAddress}`)
+    const isNftSupport =
```
kathaypacific, line 161: think this condition can be simplified to `userInfo?.appVersion ?? 0 >= '1.38.0'` (of course, pending my question on whether this `>=` actually works 😅)

src/blockscout.ts
```diff
@@ -149,13 +155,27 @@ export class BlockscoutAPI extends RESTDataSource {
       afterCursor,
     )
 
+    // For now, when you create a new transaction type other than TokenTransferV2, TokenExchangeV2
+    // You should do version check to take care of backward compatibility with wallet client.
+    const userInfo = await fetchFromFirebase(`registrations/${userAddress}`)
```
kathaypacific, line 160: can we take this version check logic into a separate function? i see it's duplicated later in this file

src/blockscout.ts
```diff
@@ -149,13 +155,27 @@ export class BlockscoutAPI extends RESTDataSource {
...
     const context = { userAddress }
 
+    // Order is important when classifying transactions.
+    // Think that below is like case statement.
     const transactionClassifier = new TransactionClassifier([
       new ExchangeContractCall(context),
       new EscrowContractCall(context),
       new ContractCall(context),
       new EscrowSent(context),
+      isNftSupport ? new NftReceived(context) : new TokenSent(context),
```
kathaypacific, line 177: when `isNftSupport` is false, `transactionClassifier` has 3 duplicate items for `TokenSent`...is this okay? is it better to default to `undefined`, and do a `.filter` to remove undefined objects from the array? i'm thinking that it's not always obvious what the fallback type should be, if we did keep using this pattern

src/blockscout.ts
```diff
@@ -230,11 +250,22 @@ export class BlockscoutAPI extends RESTDataSource {
...
+        : userInfo['appVersion'] >= '1.38.0'
+
     const supportedTokens = new Set(tokenInfoCache.getTokensAddresses())
 
     const filteredUnknownTokens = transactions.filter((tx: Transaction) => {
       return tx.transfers.every((transfer: BlockscoutTokenTransfer) => {
-        return supportedTokens.has(transfer.tokenAddress.toLowerCase())
+        return (
+          supportedTokens.has(transfer.tokenAddress.toLowerCase()) ||
+          (isNftSupport ? transfer.tokenType === 'ERC-721' : false)
```
kathaypacific, line 267: can this just be `isNftSupport && transfer.tokenType === 'ERC-721'`?

src/blockscout.ts
kathaypacific, line 161: what do you think about the name `shouldReturnNftTransactions` or `shouldIncludeNftTransactions`?

src/events/NftReceived.ts
```diff
@@ -0,0 +1,44 @@
...
+          isNftReceived = false
+        }
+      }
+    }
+
+    return (
+      transaction.transfers.length >= 1 &&
+      (numNftsSent == numNftsReceived
+        ? isNftReceived
+        : numNftsSent < numNftsReceived)
```
kathaypacific, line 30: sorry i'm still a bit confused reading this logic....what was the issue with simplifying this to something like `return numNftsReceived > 0 && numNftsReceived >= numNftsSent`? and conversely for the nft sent case, something like `return numNftsSent > 0 && numNftsSent > numNftsReceived`? (so here we use the rule where if the number of nfts sent and received is equal, we call that a "nft received")

src/firebase.ts
```diff
@@ -19,8 +19,8 @@ admin.initializeApp({
 export const database = admin.database()
 
 export async function fetchFromFirebase(path: string) {
-  const snapshot = await database.ref(path).once('value')
-  return snapshot.val()
+  const snapshot = await database?.ref(path).once('value')
```
kathaypacific, line 22: why do we need all these extra checks?

kathaypacific (commented):
Nice work!! :D

src/blockscout.ts
dievazqu: The `TransactionClassifier` receives an array, maybe you can build the array before passing it? And you can `.push` any element that meets our criteria? Also thinking this for future features, it would be nice if we have some kind of convention.

test/mockTokenTxsV2.ts
```diff
@@ -994,6 +994,182 @@ const mockTokenTxs = {
               '0xba620de2d812f299d987155eb5dca7abcfeaf154f5cfd99cb1773452a7df3d7a',
           },
         },
+        // ERC-721 transaction
```
dievazqu, line 997: Hmm, this looks good, but I don't see any change on the `blockscoutV2.test.ts` or in the mock response, I guess it's because we have to mock the client app version?
I think it would be nice if we could test a response with a NFTReceive event and with a NFTSent event. And if you really want, a test with both?

src/blockscout.ts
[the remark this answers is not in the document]
dievazqu: Did you measure how much time this request takes?
This endpoint takes some time because of the query to blockscout, so this fetch might not add a lot of time to that. But I'm wondering if we should move the fetching logic to something similar to `KnownAddressesCache` so we have all this data locally and we avoid doing the fetching in the middle of the query.
Probably we can improve this in another PR, so feel free to ignore this comment.

dievazqu (commented):
Nice work! It needs a bit of code clean up, please read Kathy's comments, and it would be great if you could add some test cases for this!

src/blockscout.ts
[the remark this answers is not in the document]
kathaypacific: ah yeah this is a good point. it would be nice to read from the firebase rtdb only once for the request and store the answer, right now we do that twice.

src/blockscout.ts
Samuel3Shin: @kathaypacific 3 duplicate items of TokenSent are okay. But I think it could be improved in other ways like you and @dievazqu's suggestions!

src/blockscout.ts
Samuel3Shin: That's more simple and concise. Thanks!

Samuel3Shin: I just changed it

src/blockscout.ts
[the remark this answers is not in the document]
Samuel3Shin: I agree with you @kathaypacific 
I just changed it to skip the check if `userAddress` is null. and make `isNftSupport` false as a default value.

And I agree you with @dievazqu, I will change it to do just once, probably next PR!

Samuel3Shin: Yes, as you and @dievazqu's opinion, I will remove this duplicate action!

src/blockscout.ts
Samuel3Shin: Oh this could be simplified to 
isNftSupport = (userInfo?.appVersion ?? '0.0.0') >= '1.38.0'

Yes, `>=` works as a string comparing.
You can test it in this web ide (https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0HQARlAAF5QAByGkBADMAE4AjSmYoKVTQDoAEwM5mstkADgCAAZecE+Y1BPB3AFoPBmAAKWmgAB8jOFAEoANzBFQ4ZoeMheRDwPoMYHsGhfNxmZjsBjMSD8WDQdSJe3wPQiLwAIgAoqsGIgFvoQ0lsSHwA5UEzhAA5SASENBFQAeS+ZHegi9WkgR1QjgUrmkeFAThtOmGWH0KAQiGhwkuzoktuiIgMHGxCyYrA4PCCJSAA)

Samuel3Shin: I changed it to `shouldIncludeNftTransactions`

src/events/NftReceived.ts
Samuel3Shin: When numNftsSent = 3, numNftsReceived = 3,
in this case, I want to check the last NFT transfer event to decide whether it is `sent` or `received`.

But in your logic, there is no checking of the last transfer event!

ps. `transaction.transfers.length >= 1` can be changed to `numNftsReceived > 0` by the way.

src/firebase.ts
Samuel3Shin: There is a mock address in the test code, the mock address causes an error when fetching RTDB from firebase!

test/mockTokenTxsV2.ts
Samuel3Shin: I see, I will add the test code with the mock client app version, if not in this PR, in the next PR!

src/blockscout.ts
```diff
@@ -149,13 +155,27 @@ export class BlockscoutAPI extends RESTDataSource {
...
 
+    // For now, when you create a new transaction type other than TokenTransferV2, TokenExchangeV2
+    // You should do version check to take care of backward compatibility with wallet client.
+    const userInfo = await fetchFromFirebase(`registrations/${userAddress}`)
+    const isNftSupport =
+      userInfo === undefined
+        ? false
+        : userInfo['appVersion'] === undefined
+        ? false
+        : userInfo['appVersion'] >= '1.38.0'
```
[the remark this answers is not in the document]
Samuel3Shin: Actually, it works you can check it on the web ide below. But I will use the library!
(https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0HQARlAAF5QAByGkBABMARpTMUFKpoB0bIZzNZNJF3OCPMagng7gC0HgzAAFLTQAA+RkCgCUAG5giocM0PGQvIh4H0GMD2DQvm4zMx2AxmJB+LBoOpEhb4HoRF4AEQAUVWDEQC30vqS2N94AcqCZwgAcpAJL6gioAPJfMjvQTOrSQI6oRwKVzSPCgJymnTDLD6FAIRDQ4SXG0SM3REQGDjYhZMVgcHhBEpAA)

src/blockscout.ts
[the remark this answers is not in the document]
Samuel3Shin: @dievazqu I just checked the time. This function takes 80ms ~ 84ms.

test/mockTokenTxsV2.ts
Samuel3Shin: I will add the test code according to your suggestion. If not in this PR, probably next PR!

src/blockscout.ts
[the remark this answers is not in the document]
Samuel3Shin: @kathaypacific I just added the module `compare-versions`, then do I need to only add `package.json`
and I don't need to add `package-lock.json` and `yarn-lock.json`. Is that right??

src/firebase.ts
kathaypacific: i don't think we should change actual code just for tests - this is a sign that the way the test is mocked is incorrect :/

src/events/NftReceived.ts
kathaypacific: right, but that is my point. why do we need to check the last transfer event? in my opinion, it should be the net result that matters. it would be less predictable if numNftsSent = 3, numNftsReceived = 3 can mean NFT set AND NFT received.

@dievazqu do you have thoughts about this? idk if i'm missing something but i don't think the order of the transfers should decide how we categorise the transaction?

src/blockscout.ts
kathaypacific: i don't think it does, even on the IDE above if you change the code to 
```
const v1 = '1.2.1'
const v2 = '1.11.0'
console.log(v1 >= v2);
```
the outcome is "true" but 1.2.1 is clearly not higher than 1.11.0

src/blockscout.ts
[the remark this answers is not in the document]
kathaypacific: no, you should also check in the lock files. an explanation is [here](https://11sigma.com/blog/2021/09/03/yarn-lock-how-it-works-and-what-you-risk-without-maintaining-yarn-dependencies-deep-dive/#:~:text=lock%20is%20the%20main%20source,a%20project%20(listed%20in%20yarn.) 

and to convince yourself that the `compare-versions` is needed, you can try the following in the editor you mentioned above
```
const v1 = '1.2.1'
const v2 = '1.11.0'
console.log(v1 >= v2);
```
it will tell you true :) i guess we are fine for the case where we use the same number of digits for each part of the version, but this is pretty brittle and could cause unexpected issues down the line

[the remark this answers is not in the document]
dievazqu: Can we add a small test to be sure it works as we expect? 😅

src/blockscout.ts
```diff
@@ -149,13 +156,35 @@ export class BlockscoutAPI extends RESTDataSource {
       afterCursor,
     )
 
+    // For now, when you create a new transaction type other than TokenTransferV2, TokenExchangeV2
+    // You should do version check to take care of backward compatibility with wallet client.
+
+    let shouldIncludeNftTransactions = false
```
kathaypacific, line 162: hum, so i still see this logic being duplicated twice. can we do this one time, to avoid hitting the db unnecessarily? you just never know when this db fetch could be unexpectedly slow, and i don't think we actually need to do it twice. 

one option is to move this block to the top of `getTokenTransactionsV2`, and pass the result in to `getRawTokenTransactionsV2`. i'm not very familiar with blockchain-api but @dievazqu also mentioned the `KnownAddressCache` which might achieve a similar effect of querying the db only once

src/events/NftReceived.ts
dievazqu: I'm not sure how to categorise transactions with same amount of nfts received and sent.
Do we have real cases with that situation?
For me, those situations sound like a swap of NFTs, maybe we should create a `NFT swap` event?
In that case, I would leave the `NFT received` when we don't have any `NFT sent`, and vice versa.

src/firebase.ts
dievazqu: I agree with Kathy, I think you can mock the `fetchFromFirebase` call.

test/mockTokenTxsV2.ts
dievazqu: I would appreciate if you can add the test in this PR, since I think it's really important that we have the expected behaviour, so we don't end up in another backward compatibility issue. I wouldn't like this to be deployed without those tests.

kathaypacific: actually @dievazqu i think this did change the tests, but the tests didn't complain because of this update https://github.com/valora-inc/blockchain-api/pull/182/files#r936430749

src/firebase.ts
kathaypacific: one dumb way i can think of doing it is something like 
```
jest.mock('firebase-admin', () => ({
  initializeApp: jest.fn(),
  database: () => ({
    ref: () => ({
      once: () => ({
        val: jest.fn(),
      }),
    }),
  }),
}))
```
in the blockscountV2.test.ts file....there will be other ways, but yeah i think we should focus on getting our mocks right rather than changing our code.

src/blockscout.ts
kathaypacific: i think having 3 of the TokenSent items means that you go into the checks that the class contains, 3 times? i guess this works but it's not optimal 😅

src/events/NftReceived.ts
kathaypacific: haha ah, complexity 🙈 i think lets keep NFT swaps out of scope for now. 

but i guess this is kind of my point - we should focus on the net number of received vs sent, rather than the order right? honestly i don't know how likely it is that the net number is non zero AND equal, but if this happens i'd expect it to be attributed to one outcome (not 2). so going forward, if we do want to say that this is a swap, then we only need to move this type of transaction from the NFT received or NFT sent transaction rather than both?

src/firebase.ts
dievazqu: I think it's simpler to mock the `firebase.ts`, something like:

```
const mockFetch = jest.fn()
jest.mock('../src/firebase.ts', () => ({
  fetchFromFirebase: (path: string) => mockFetch(path)
}))
```

Then, whenever you want to define the mock value you can do:
```
mockFetch.mockImplementation((path: string) => {
  return mockValue
})
```

But Kathy's way also works.

src/events/NftReceived.ts
dievazqu: Yeah, definitely it's out of scope 🤣 
I agree that the order shouldn't change the category of the event, so I guess we have to decide if we want to show `Sent` or `Receive` when the amount is equal... Maybe it's not the most popular option, but for me, if the transaction has the same amount, I would ignore it... I would only show those which have more sent or received nfts.
But any decision on this seems fine, I don't think we will have a lot of those cases anyways

src/firebase.ts
Samuel3Shin: Oh, the test codes are already written, not what I added!

Do I need to change the already written test codes??

src/blockscout.ts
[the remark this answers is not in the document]
Samuel3Shin: Yup @kathaypacific You're right! I will use the `compare-versions` module.

And I added `package.json` and `yarn.lock`.

I didn't add `package-lock.json` because this repo doesn't have `package-lock.json` and I think it is auto-created when we do `yarn`. Is that correct?? Or do I need to add `package-lock.json` as well??

src/blockscout.ts
Samuel3Shin: @kathaypacific You're right! I will use the module `compare-versions`!

src/blockscout.ts
```diff
@@ -149,13 +156,35 @@ export class BlockscoutAPI extends RESTDataSource {
...
       new ExchangeContractCall(context),
       new EscrowContractCall(context),
       new ContractCall(context),
       new EscrowSent(context),
+      shouldIncludeNftTransactions
+        ? new NftReceived(context)
+        : new TokenSent(context),
+      shouldIncludeNftTransactions
+        ? new NftSent(context)
+        : new TokenSent(context),
```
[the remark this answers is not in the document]
Samuel3Shin: I'm now trying to make it `undefined` and filter it later.
But the error message shows that "Type 'undefined' is not assignable to type 'TransactionType'."

Do you think it is safe to add `undefined` in the `TransactionType`??

Screen Shot 2022-08-04 at 12 46 51 PM


Screen Shot 2022-08-04 at 12 47 02 PM

src/blockscout.ts
Samuel3Shin: Yup, I removed the duplicate function call in the way you told me. I passed the result to `getRawTokenTransactionsV2`

src/firebase.ts
Samuel3Shin: It is these three mock addresses that made me check the `fetchFromFirebase()`.
I can just change the mock addresses to the real one that the Firebase RTDB has.
What should be a proper way to handle this situation???

src/events/NftReceived.ts
Samuel3Shin: I changed it to @kathaypacific's proposal!

src/blockscout.ts
Samuel3Shin: I removed the duplicate action!

src/blockscout.ts
Samuel3Shin: I'm now trying to make it `undefined` and filter it later.
But the error message shows that "Type 'undefined' is not assignable to type 'TransactionType'."

Do you think it is safe to add `undefined` in the `TransactionType`??

Screen Shot 2022-08-04 at 12 46 51 PM


Screen Shot 2022-08-04 at 12 47 02 PM

test/mockTokenTxsV2.ts
Samuel3Shin: I'm working on it now. I will add both `sent` and `received` events and I will also add test codes for the version check as well.

Samuel3Shin: I added test codes for the two events and a wallet version check.

src/blockscout.ts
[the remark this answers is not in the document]
Samuel3Shin: @dievazqu I added test codes for wallet versions!

yarn.lock
```diff
@@ -4,7 +4,7 @@
 
 "@apidevtools/json-schema-ref-parser@^9.0.3":
   version "9.0.9"
-  resolved "https://registry.yarnpkg.com/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-9.0.9.tgz#d720f9256e3609621280584f2b47ae165359268b"
+  resolved "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-9.0.9.tgz"
```
kathaypacific, line 7: hey did you do a `yarn` or a `npm install`? i don't think all these changes for the registry are right, there should only be changes related to `compare-version` here

kathaypacific:
p.s. i set up blockchain-api locally and tested the backwards compatibility + against your wallet branch and both work great!

yarn.lock
Samuel3Shin: `npm install` caused that change. But I re-install with yarn!

test/blockscoutV2.test.ts
```diff
@@ -98,6 +98,23 @@ const mockCurrencyConversionAPI: CurrencyConversionAPI = {
...
+  fetchFromFirebase: (path: string) => mockFetch(path),
+}))
+
+mockFetch.mockImplementation((path: string) => {
+  return {
+    appVersion: '1.38.0',
+    country: 'US',
+    dailyLimitCusd: 1.7976931348623157e308,
+    fcmToken:
+      'eiyKgZwIT528ekhKOxwdKj:APA91bEXvrKfCGjOGPhuEf2DjytINQK7nnuVnD5ez4QsagTzl2mfOZhYAw0JXTyWiTYOt2723M0X67yzI4Mn_HChZfcKmoyoE31iYNNWUf7af9cSU-fuQBpZeyHOv3Fv3IsIMY1Z5nbF',
```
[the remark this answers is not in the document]
Samuel3Shin: I thought real value would be good for the testing because it is real value lol.
Anyway, I changed it to fake value.

src/blockscout.ts
```diff
@@ -206,6 +244,7 @@ export class BlockscoutAPI extends RESTDataSource {
 
   async getRawTokenTransactionsV2(
     address: string,
+    isIncludeNftTransactions: boolean,
```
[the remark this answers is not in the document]
Samuel3Shin: I changed it!

src/blockscout.ts
```diff
@@ -144,25 +151,56 @@ export class BlockscoutAPI extends RESTDataSource {
...
+    const filteredClassifyingCandidates = []
+
+    for (const eventType of classifyingCandidates) {
+      if (eventType !== undefined) {
+        filteredClassifyingCandidates.push(eventType)
+      }
+    }
+
+    const transactionClassifier = new TransactionClassifier(
+      filteredClassifyingCandidates,
```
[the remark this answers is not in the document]
Samuel3Shin: Thanks! I changed my codes to this!

test/blockscoutV2.test.ts
```diff
@@ -98,6 +98,23 @@ const mockCurrencyConversionAPI: CurrencyConversionAPI = {
   getFromMoneyAmount: jest.fn(),
 }
 
+const mockFetch = jest.fn()
+jest.mock('../src/firebase.ts', () => ({
+  fetchFromFirebase: (path: string) => mockFetch(path),
+}))
+
+mockFetch.mockImplementation((path: string) => {
```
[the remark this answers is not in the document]
Samuel3Shin: If I don't do global mocking for `fetchFromFirebase()`, I need to change the function again because already written test codes cause a problem while calling `fetchFromFirebase()`.

from
Screen Shot 2022-08-05 at 8 46 14 AM

to
Screen Shot 2022-08-05 at 8 46 32 AM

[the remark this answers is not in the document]
Samuel3Shin: I can make the version less than 1.38.0 in the global mocking if I don't want to change the snapshot.

test/__snapshots__/blockscoutV2.test.ts.snap
```diff
@@ -398,6 +398,18 @@ Array [
     "transactionHash": "0x34e6e74bc01c7112817e669a8057ae7f4c1ed49d8de824bea8ecbdb945b41345",
     "type": "SENT",
   },
+  Object {
```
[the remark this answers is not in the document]
Samuel3Shin: I changed the version of global mocking not to change the snapshot.

test/blockscoutV2.test.ts
```diff
@@ -133,3 +150,62 @@ describe('Blockscout', () => {
...
+    const stringBuilder: string[] = []
+
+    for (const transaction of result.transactions) {
+      stringBuilder.push(JSON.stringify(transaction))
+    }
+
+    const resultString = stringBuilder.join('')
+
+    expect(resultString).toEqual(expect.not.stringContaining('NFT_SENT'))
+    expect(resultString).toEqual(expect.not.stringContaining('NFT_RECEIVED'))
```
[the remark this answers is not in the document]
Samuel3Shin: I don't think that this test will pass for the mainnet incident.

This is mainnet incident flow.
1. `blockchain-api` made all the responses
2. `wallet` repo doesn't call the `NFT events` in the graphQL.
3. The responses of NFT events become a blank in the graphQL.

In this test code, it checks the 1. `blockchain-api` made all the responses.
It checks whether the `blockchain-api` created the `NFT events` or not.
You can verify this with printing ` blockscoutAPI.getTokenTransactionsV2()`


I already tried your approach... So I tried to find the `eventType` that I can see in the console log of the result. But there was no `eventType` like the below screenshot.
Screen Shot 2022-08-05 at 9 03 10 AM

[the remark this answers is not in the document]
Samuel3Shin: I already tried your approach... So I tried to find the `eventType` that I can see in the console log of the result. But there was no `eventType` like the below screenshot.

Screen Shot 2022-08-05 at 9 03 10 AM

test/blockscoutV2.test.ts
```diff
@@ -133,3 +150,62 @@ describe('Blockscout', () => {
...
+    const stringBuilder: string[] = []
+
+    for (const transaction of result.transactions) {
+      stringBuilder.push(JSON.stringify(transaction))
+    }
+
+    const resultString = stringBuilder.join('')
+
+    expect(resultString).toEqual(expect.stringContaining('NFT_SENT'))
+    expect(resultString).toEqual(expect.stringContaining('NFT_RECEIVED'))
```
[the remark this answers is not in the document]
Samuel3Shin: Answered in the upper thread!

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
kathaypacific: thanks! the tokens are sensitive, so we should never use real values and check them in to a public project. we weren't checking against the value in any test here either so i do think it's better to have a mock value

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
kathaypacific: oh i meant removing only this `mockImplementation` - we can still keep the `jest.mock('../src/firebase.ts', () => ({...})`, but we don't really need to provide global default value for the `mockFetch`, i think the tests still pass like this. this is the shorter way to change the value of the `appVersion` to 1.36 like you've done, its okay to leave it like this

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
kathaypacific: ooohhhh okay sorry i misunderstood how this works. you're right, this is fine then. can i suggest simplifying the string builder thing to `const resultString = JSON.stringify(result)`? should achieve the same result with less loops :)

kathaypacific (approved):
just had one small suggestion for the test, but looks good to me!

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
Samuel3Shin: Oh, thanks for the tip! I changed it!

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
Samuel3Shin: I see. I will keep that in my mind!

test/blockscoutV2.test.ts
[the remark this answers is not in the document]
Samuel3Shin: I see. Sorry, I misunderstood your intention. I just removed `mockImplementation` and it works well. Thanks!

PR #183, opened 2022-08-03
feat(in-app swaps): support swap events

Samuel3Shin:
### Description
Added Swap transaction event to support swap transaction home feed.

The algorithm to find the swap transaction. (Please look at `isSwap()` in `TransfersUtils.ts`) (Thanks @kathaypacific!)
It is a swap transaction if there is 1 transfer that is "from" the user and 1 transfer that is "to" the user in its transfers.
If the transaction is not in this format, we don't create a swap event even though the transaction was created by some swap Dapp.

Currently, we support `Ubeswap`, `Symmetric`, `Mobius`, `Mento-Fi`.

We don't support `Uniswap` because the swap transaction from Uniswap has more than 1 transfer that is "from" the user and more than 1 transfer that is "to" the user. (This might be next feature that we need to work on)

home feed <--------------------------------------------> transaction detail
Screen Shot 2022-08-12 at 1 37 16 PM Screen Shot 2022-08-03 at 8 54 50 AM 

### Other changes
N/A

### Tested
With Mac Pro 2019

### How others should test
With local machine

### Related issues
- https://github.com/valora-inc/wallet/pull/2752
- https://linear.app/valora/issue/RET-290/add-swaps-to-transaction-feed

### Backwards compatibility
Will add version check after NFT event PR is merged


src/blockscout.ts
```diff
@@ -189,6 +197,9 @@ export class BlockscoutAPI extends RESTDataSource {
         new TokenReceived(context),
         new ExchangeCeloToToken(context),
```
kathaypacific, line 198: can we add the swap transactions classifier above this one? then for app versions > 1.39, we'd stop showing the custom items for CELO right? i don't think we need to make the distinction of a swap vs a swap with CELO anymore (but we shouldn't remove `ExchangeCeloToToken` and `ExchangeTokenToCelo` for backwards compatibility

src/events/SwapTransaction.ts
```diff
@@ -0,0 +1,42 @@
+import { isSwap } from '../transaction/TransfersUtils'
+import { EventBuilder } from '../helpers/EventBuilder'
+import { Transaction } from '../transaction/Transaction'
+import { TransactionType } from '../transaction/TransactionType'
+import { TokenTransactionTypeV2 } from '../resolvers'
+
+export class SwapTransaction extends TransactionType {
+  matches(transaction: Transaction): boolean {
+    return (
+      transaction.transfers.length >= 2 &&
```
kathaypacific, line 10: i think you don't need this one right, because `isSwap` would already return false if this condition is false

src/events/SwapTransaction.ts
```diff
@@ -0,0 +1,42 @@
...
+      isSwap(transaction.transfers, this.context.userAddress) &&
+      transaction.transfers.every((transfer) => {
+        return transfer.tokenType === 'ERC-20'
+      })
+    )
+  }
+
+  async getEvent(transaction: Transaction) {
+    const inTransfer = transaction.transfers[transaction.transfers.length - 1]
+    const outTransfer = transaction.transfers[0]
```
kathaypacific, line 20: i think relying on the order of transfers is not going to work in every scenario (e.g. https://explorer.celo.org/tx/0x2e3a03679d5d5ea16cfdf1321a5f2d48edbbc5ea8c667ce3b884e1dd97aa0f48/token-transfers) 

in this case, we should define the `inTransfer` as the transfer that has the user's address as the "to" address, and similarly for `outTransfer`?

kathaypacific (commented):
Looking great! it'd be cool to have a test for this too :)

src/transaction/TransfersUtils.ts
```diff
@@ -137,3 +137,23 @@ export function popTransferTo(
       transfer.toAddressHash.toLowerCase() === recipient,
   )
 }
+
+export function isSwap(
+  tokenTransfers: BlockscoutTokenTransfer[],
+  userAddress: string,
+): boolean {
+  if (tokenTransfers[0].fromAddressHash.toLowerCase() !== userAddress)
```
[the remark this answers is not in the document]
Samuel3Shin: I also thought about that algorithm. But that would have some false positives (classified as a swap event even though it is not a swap event). Of course, it will show some wrong value in the swap feed item and detailed transaction.

But I'm not sure about how many false positives would be in the raw transactions. I'm willing to change the algorithm to that by the way.

src/blockscout.ts
Samuel3Shin: Yes, I will do that!

src/events/SwapTransaction.ts
Samuel3Shin: If a transaction has one transfer and its `to address` and `from address` are both the `user's address`, then it will pass `isSwap()`.
I'm not sure whether it exists in the raw transactions or not though.
I can remove this part if we are sure that there is no transaction like that!

src/events/SwapTransaction.ts
Samuel3Shin: Actually, some swap transaction has 2 `inTransfer` to `user's address` and 2 `outTransfer` to `user's address'  in one swap transaction. 

In current `isSwap()` algorithm, we just skip it. But if we one to change the algorithm to "it is a swap if in all the transfers, there is at least 1 transfer that is "from" (but not "to") the user and at least 1 transfer that is "to" (but not "from") the user.", we need to aggregate all the `inTransfers` to make one `inTransfer` and aggregate all the `outTransfers` to make one `outTransfer`.

src/events/SwapTransaction.ts
kathaypacific: hmm i think that the `isSwap` function should always try to find at least 2 transfers, 1 to the user and 1 from the user right? it shouldn't return true if it's just one transfer to **and** from the user.

src/events/SwapTransaction.ts
kathaypacific: yeah that's a good point, maybe we should ignore that scenario for now and only surface swaps that have 1 transfer to the user and 1 transfer from the user. 

i'm not sure how common it is to have more than 1 transfer to and from the user, but i think it makes sense to leave that as a follow up feature if we see that it is necessary

src/events/SwapTransaction.ts
Samuel3Shin: Yes, I can put the logic inside of the isSwap(). I will remove the `transaction.transfers.length >= 2` condition!

src/events/SwapTransaction.ts
Samuel3Shin: I see! By ignoring that situation, your algorithm "it is a swap if in all the transfers, there is at least 1 transfer that is "from" (but not "to") the user and at least 1 transfer that is "to" (but not "from") the user.", works a charm. I will keep this algorithm!

src/transaction/TransfersUtils.ts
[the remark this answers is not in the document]
Samuel3Shin: I tested with my transactions below. Your proposed algorithm works so well. I will use this algorithm! I think there would not be many false positives if we add the condition that "only 1 `inTransfer` and 1 `outTransfer` should be in the transaction."

`// Transaction Log Ubeswap, Symmetric, Mobius
Celo -> cBTC (Symmetric): 0xd4ab42f041aeb88f90fab4fb3a85afee3a8fdc7f3f865df58fc026a052da80b9
Celo -> cUSD (Symmetric): 0x3b77d6df1eb26c476ff606bc79fd213f58d2b0c5bdddd7a906f4cf8a1eded2e7
Celo -> UBE (Symmetric): 0xe2cd0698b93b24540cc45b25a23c65e707bc95ee625a56481f0c54588e39a4df
FTM -> BNB (Ubesawap): 0x5122a8ad1ae48f50ac886b736a01287a1669348b40cbcd5b19a98d3a238048f5
Celo -> cEUR (Mento-Fi): 0x39f390d14dbc0887cc933e0ee03567b7bb592e0b905e4c95347c0e1a81b64a22
Celo -> cUSD (Mento-Fi): 0x228a99d2b7d9123dec35d235eff452904d8d485a2def8a216e0a5c0ab9a518b1
Celo -> FTM (Ubeswap): 0xf958663fe998d48da8623849972b0c19aa8b7d8c7607fa985a8ea67e2ed520da
Celo -> AVAX (Ubeswap): 0xcdf4dd9bfe4e0c76df47cd8c5d297dc7e9ac84aa7eb37d429a365f3056059c60
Celo -> BNB (Ubeswap): 0x483992b9f80831511b497f99c006564746a23b3d77381eb0f6fb4735b6181707
cREAL -> cETH (Ubeswap): 0xba254813e1aa7dcdf58412fb17f3b4e49819a67e7e5896d41995f05b2e915793
wallet->Metamask: 0x26c4d78c88ae5e3d59d360162a5e7d9f88c49f123d3a28c74dc08396cd1dcfd2
Celo->cREAL (Mento-Fi): 0x8964fb3c1221d3d4197b81857267e4a2aa5f08913fe4e69b5c87fcdf89b2c8aa
Celo->cUSD (Symmetric): 0x0c7d7b3552ffaf9efd725d1f38db71c6dc732198655e4a77b49f978169816dd6
Celo->cUSD: 0x0c0ee6decd70b8b67b12ed3d0e5e314e9c7d0e1edc7352be043b8e283870a17f
Receiving cEUR: 0xd669c9487fcba6ce96cc3f292037d4e11a8319908c185d127245373a964b6b8e
Celo->cEth (Mobius): 0x34b956535baac6a8978566b7559f0c98dfe6453faae18fd2b86f451a27ff7ff8
Celo->Ube (Ubeswap): 0xddbdee4f2aa7ca7fd7b70fda0d6ea06fb58f8a0f43b9aae82c3efa21e7eb97d9
Upbit->wallet: 0x120125dbf4979a7a91d8efa13de7660ee3bdffef4a93cef524fe30a217e92563`

kathaypacific (approved):
i think this looks really great :) i tested locally and all my swaps are being displayed correctly, it looks awesome!

i know you're worried about how robust the function for detecting swaps is, but i think how it is now is a great start and works for most use cases. it's easy for us to update this as we discover new cases since it is a backend change, so i think we don't need to feel too much pressure to get it right for 100% of cases (i don't even think that's possible at this stage 😅)

PR #185, opened 2022-08-04
feat: add the ability to manually deploy from main

MuckT:
### Description
Adds a manual dispatch to the workflow and another new workflow to manually dispatch main to Alfajores.
We could remove the check to allow branches to be deployed to Alfajores, but this could mess with CI as E2E tests for the wallet currently target Alfajores.


MuckT:
Closing as rerunning previous successful deploys meets this need and Alfajores should always match main. If developers need to test it should be done with a local instance of blockchain-api instead of a deployed version of Alfajores.

PR #217, opened 2022-11-09
fix: support new apilayer currency data API

jeanregisser:
### Description

Temporary change, the new account I created for alfajores can only use the new API.
Mainnet still uses the legacy account.

### Tested

Locally

### Backwards compatibility

Yes

### Related issues

[Slack thread](https://valora-app.slack.com/archives/C025V1D6F3J/p1667992388183679).


kathaypacific (approved):
🚀

jeanregisser:
Merging now to unblock the wallet CI. But happy to follow up later with any suggestions.

PR #225, opened 2023-01-06
feat: use Valora version when present in the User-Agent header

jeanregisser:
### Description

Now that Valora will start sending the app version in the `User-Agent` header, we can avoid fetching it from Firebase when present.

### Tested

Updated tests

### Backwards compatibility

Yes

### Related issues

- Fixes RET-558


dievazqu (approved):
LGTM

silasbw (approved):

PR #234, opened 2023-02-24
fix: set min liquidity to 10k to get cREAL price back

jeanregisser:
### Description

cREAL liquidity dropped bellow $20k on 2023-02-16 which caused the cREAL price to become unavailable in Valora.

This PR decreases the min liquidity to $10k to fix this.

See Slack [thread](https://valora-app.slack.com/archives/C029Z1QMD7B/p1677235014387249?thread_ts=1677089358.452749&cid=C029Z1QMD7B).

### Tested

cREAL price becomes available again.

### Backwards compatibility

Yes

### Related issues

N/A

PR #235, opened 2023-02-24
fix: only set min liquidity for mainnet

jeanregisser:
### Description

Following up on https://github.com/valora-inc/blockchain-api/pull/234 to only set the min liquidity for mainnet.
Alfajores already has a [low min liquidity set to 100](https://github.com/valora-inc/exchanges/blob/496e20a758ea50296ae8d48ffdc49c8ffd041e3c/src/config.ts#L46).


dievazqu (approved):
LGTM

PR #249, opened 2023-05-16
feat: optionally return nft metadata and media

MuckT:
### Description

Adds the `tokenId` the blockscout query which in combination with the `tokenAddress` allows us to lookup the nft metadata and media. A request for `/getNft` is sent for each nft in the transaction where the `toAddressHash` matches the `ownerAddress` if `TokenTransactionTypeV2.NFT_RECEIVED` or `fromAddress` in the case of   `TokenTransactionTypeV2.NFT_SENT`.

### Tested

Tested locally with a local instance of blockchain-api

### Backwards compatibility

Yes - both the existing an new queries will return valid results.

### Related issues

- [RET-594](https://linear.app/valora/issue/RET-594/[blockchain-api]-return-nft-metadata-in-the-homefeed)
- https://github.com/valora-inc/nft-indexer/pull/318
- https://github.com/valora-inc/nft-indexer/pull/319
- https://github.com/valora-inc/wallet/pull/3800

#### Sample Query
```GraphQL
{ 
    tokenTransactionsV2(
    address: "0xde33e71faecdead20e6a8af8f362d2236cba005f",
    localCurrencyCode: "USD") {
        transactions {
            type
            block
            transactionHash
            ... on NftTransferV2 {
                nfts {
                    tokenId
                    contractAddress
                    tokenUri
                    ownerAddress
                    metadata {
                        id
                        name
                        description
                        image
                        dna
                        date
                    }
                    media {
                        raw
                        gateway
                    }
                }
            }
        }
    }
}
```

#### Sample Response
```GraphQL
{
    "data": {
        "tokenTransactionsV2": {
            "transactions": [
                ...
                [
                  {
                    "type":"SENT",
                    "block":"19291924",
                    "transactionHash":"0x68a21ce0c423ee22620b9be29278b5169645471b1b3d8a2d18e5337835854ae9"
                  },
                  {
                    "type":"NFT_RECEIVED",
                    "block":"19253934",
                    "transactionHash":"0x062e3514114a1641bb74460c5a98453809ed1486fe7039da71020301cc32b97a",
                    "nfts":[
                       {
                          "tokenId":"6293",
                          "contractAddress":"0xac80c3c8b122db4dcc3c351ca93ac7e0927c605d",
                          "tokenUri":"https://ipfs.io/ipfs/QmTn1W5CpTdqrkvdSLb7nXGWVYYmoTWMv8N2ripQthXw2v/6293.json",
                          "ownerAddress":"0xde33e71faecdead20e6a8af8f362d2236cba005f",
                          "metadata":{
                             "id":6293,
                             "name":"Celostrial #6293",
                             "description":"The worlds first nfETs",
                             "image":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6293.png",
                             "dna":"2894cac50a25044ef04cc3058bb9273f71725763",
                             "date":1644389874737
                          },
                          "media":[
                             {
                                "raw":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6293.png",
                                "gateway":"https://cloudflare-ipfs.com/ipfs/Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6293.png"
                             }
                          ]
                       },
                       {
                          "tokenId":"6294",
                          "contractAddress":"0xac80c3c8b122db4dcc3c351ca93ac7e0927c605d",
                          "tokenUri":"https://ipfs.io/ipfs/QmTn1W5CpTdqrkvdSLb7nXGWVYYmoTWMv8N2ripQthXw2v/6294.json",
                          "ownerAddress":"0xde33e71faecdead20e6a8af8f362d2236cba005f",
                          "metadata":{
                             "id":6294,
                             "name":"Celostrial #6294",
                             "description":"The worlds first nfETs",
                             "image":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6294.png",
                             "dna":"77c251ffa01132e1cf747afd6f7da93b3a37a6ae",
                             "date":1644389403533
                          },
                          "media":[
                             {
                                "raw":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6294.png",
                                "gateway":"https://cloudflare-ipfs.com/ipfs/Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6294.png"
                             }
                          ]
                       }
                    ]
                  },
                  {
                    "type":"NFT_RECEIVED",
                    "block":"19253763",
                    "transactionHash":"0x455c4c45159b793e7d678fe9e49feecac2a3a339df68fa4eb16a790bdee62729",
                    "nfts":[
                       {
                          "tokenId":"6295",
                          "contractAddress":"0xac80c3c8b122db4dcc3c351ca93ac7e0927c605d",
                          "tokenUri":"https://ipfs.io/ipfs/QmTn1W5CpTdqrkvdSLb7nXGWVYYmoTWMv8N2ripQthXw2v/6295.json",
                          "ownerAddress":"0xde33e71faecdead20e6a8af8f362d2236cba005f",
                          "metadata":{
                             "id":6295,
                             "name":"Celostrial #6295",
                             "description":"The worlds first nfETs",
                             "image":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6295.png",
                             "dna":"1f86a290e86bf9dd01c2144fd327b3de14811954",
                             "date":1644389630153
                          },
                          "media":[
                             {
                                "raw":"ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6295.png",
                                "gateway":"https://cloudflare-ipfs.com/ipfs/Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6295.png"
                             }
                          ]
                       }
                    ]
                  }
                ...
            ]
        }
    }
}
```


MuckT:
@kathaypacific think I just need some tests and this is good to go. 🙏

src/schema.ts
```diff
@@ -172,12 +172,35 @@ export default `type ExchangeRate {
...
+
   type NftTransferV2 implements TokenTransactionV2 {
     type: TokenTransactionTypeV2!
     transactionHash: String!
     timestamp: Timestamp!
     block: String!
     fees: [FeeV2]
+    tokenId: String
+    tokenAddress: Address
+    nft: Nft
```
kathaypacific, line 203: any reason why you want to put the tokenUri and metadata inside an object? why not have both on the top level?

src/helpers/EventBuilder.ts
```diff
@@ -65,12 +67,27 @@ export class EventBuilder {
     const transactionHash = transaction.transactionHash
     const block = transaction.blockNumber
     const timestamp = transaction.timestamp
+    // We return the first NFT found in the transaction transfers
```
kathaypacific, line 70: the current logic for "nft received" is just that the number of incoming nft transfers in the transaction is greater than the number of outgoing nft transfers. (vice versa for "nft sent"). for transactions that have more than 1 nft transfer, returning the first NFT might be incorrect. 

what do you think about also including a check using the `eventType`, so the condition is like "find the first nft that is going to / from the user's wallet"?

src/config.ts
```diff
@@ -54,3 +54,5 @@ export const STAKED_CELO_TOKEN_ADDRESS = (
 export const STAKED_CELO_MANAGER_ADDRESS = (
   process.env.STAKED_CELO_MANAGER_ADDRESS as string
 ).toLowerCase()
+
+export const CLOUD_FUNCTIONS_URL = process.env.CLOUD_FUNCTIONS_URL
```
kathaypacific, line 58: i think you'll need to add this CLOUD_FUNCTIONS_URL to the alfajores / mainnet yaml files?

aside, we haven't been very good about conventions but i saw silas recently bringing this up https://github.com/valora-inc/wallet/pull/3730#discussion_r1190070464 maybe we also want to use that here?

kathaypacific (commented):
looking great Tom!!!

src/schema.ts
MuckT: I think this format is a better representation of the data. We've made the choice to only return 1 nft per transaction knowing that this is not always true, if we add `tokenUri` & `metadata` we build further on this. 

It seems like this format would extend a better to multiple Nfts if we added an `nfts: nft[]`. It also looks more graphQL'y for lack of a better term. Screenshot 2023-05-16 at 6 36 42 PM

src/helpers/EventBuilder.ts
MuckT: This should be fixed by 563a4e4062b0639e8767de19d28883473272d97a.

src/helpers/EventBuilder.ts
```diff
@@ -65,12 +67,27 @@ export class EventBuilder {
...
+    const tokenId =
+      transaction.transfers.find((transfer) => transfer.tokenId)?.tokenId ??
+      null
+    const tokenAddress =
+      transaction.transfers.find((transfer) => transfer.tokenId)
+        ?.tokenAddress ?? null
```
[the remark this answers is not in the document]
MuckT: Fixed in 563a4e4062b0639e8767de19d28883473272d97a!

src/config.ts
silasbw: +1 -- how about calling this GET_NFT_API_URL and setting it to api.alfajores.valora.xyz/getNft or api.mainnet.valora.xyz/getNft?

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +166,33 @@ export class EventBuilder {
...
+        tokenId,
+        contractAddress,
+        ownerAddress: null,
+        tokenUri: null,
+        metadata: null,
+      }
+    }
+    const { result } = await response
+      .json()
+      .catch((err) =>
```
silasbw, line 185: what's up with the `catch` here?  why not use try/catch?

src/helpers/EventBuilder.ts
```diff
@@ -58,19 +61,52 @@ export class EventBuilder {
...
+            transfer.fromAddressHash === address
+        : (transfer: { toAddressHash: string | null }) =>
+            transfer.toAddressHash === address
+
+    const filteredTransfers = transaction.transfers.filter(filterFn)
+
+    // Use Promise.all to get all NFTs in parallel
+    nfts = await Promise.all(
+      filteredTransfers.map(async ({ tokenId, tokenAddress }) => {
+        return await EventBuilder.getNft(tokenId, tokenAddress)
```
silasbw, line 96: How many tokens do you expect to fetch here?  Should you set a limit on the concurrency?

src/helpers/EventBuilder.ts
MuckT: Since the return type is `Promise<Nft>` when I tried try/catch I ended up having to throw from the catch block to keep the correct return types. I opted to use .catch logger errors instead to keep the return types and at minimum return the contractAddress and tokenId from BlockScout

src/helpers/EventBuilder.ts
MuckT: If there are multiple tokens from a booster pack like setting it could 10 - 16 nfts. This would be the higher end, more than likely it would be one or two. Does 10 seem reasonable for now?

MuckT: Added a concurrency limit of 5 in 015bea68fc2ebdd3863d249d142a833f464d96eb.

src/events/NftReceived.ts
```diff
@@ -3,19 +3,20 @@ import { TokenTransactionTypeV2 } from '../resolvers'
 import { Transaction } from '../transaction/Transaction'
 import { TransactionType } from '../transaction/TransactionType'
 
+let address: string | null = null
```
dievazqu, line 6: I think we don't need this variable, you can just use `this.context.userAddress.toLowerCase()` where we need.
If you still want to use an extra variable, it shouldn't be at the global scope, move it inside the class.
If not, creating `NftReceived` events from different address can bring inconsistencies (assigning the wrong address).

src/events/NftSent.ts
```diff
@@ -3,19 +3,20 @@ import { TokenTransactionTypeV2 } from '../resolvers'
 import { Transaction } from '../transaction/Transaction'
 import { TransactionType } from '../transaction/TransactionType'
 
+let address: string | null = null
```
dievazqu, line 6: Same here.

src/helpers/EventBuilder.ts
```diff
@@ -58,19 +61,58 @@ export class EventBuilder {
...
+    // Top check if the event type is valid
+    if (
+      eventType !== TokenTransactionTypeV2.NFT_SENT &&
+      eventType !== TokenTransactionTypeV2.NFT_RECEIVED
+    ) {
+      throw new Error(`Invalid event type ${eventType}`)
+    }
+    
+    // Filter by address based on the event eventType - fromAddress for sent, toAddress for received
+    const filterFn =
```
dievazqu, line 84: nit: I think it's still unclear how we wanna handle multiple NFTs. So we can revisit this later.
But, I think it might make sense to include all the NFTs transfers that are related to the user address, all the in and out going transfers no matter the type. So, if there's like a swap we will have the info for all the NFTs swapped.

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +172,33 @@ export class EventBuilder {
...
+        tokenUri: null,
+        metadata: null,
+      }
+    }
+    const { result } = await response
+      .json()
+      .catch((err) =>
+        logger.error(`Failed to parse response from ${url} - ${err}`),
+      )
+    const { tokenUri, metadata, ownerAddress } = result
```
dievazqu, line 194: Won't this fail if `result` is `undefined`? Which could happen if parsing to json has failed, right?
I think we should return this inside the `catch`:
```
return {
  tokenId,
  contractAddress,
  ownerAddress: null,
  tokenUri: null,
  metadata: null,
}
```

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +172,33 @@ export class EventBuilder {
   static getWeiForToken(address: string) {
     return Math.pow(10, tokenInfoCache.getDecimalsForToken(address))
   }
+
+  static async getNft(tokenId: string, contractAddress: string): Promise<Nft> {
+    const url = `${GET_NFT_API_URL}?contractAddress=${contractAddress}&tokenId=${tokenId}`
+    const response = await fetch(url)
```
dievazqu, line 178: Checking on the example [here](https://www.npmjs.com/package/cross-fetch), I think this should be wrapped inside a `try/catch` block.
Maybe you can put the whole function code inside a `try/catch` block, and return the "default" object (the one with the nulls) in case of an error.

src/schema.ts
```diff
@@ -172,12 +172,37 @@ export default `type ExchangeRate {
     account: String!
   }
 
+  type NftAttributes {
+    trait_type: String
+    value: String
+  }
+
+  type NftMetadata {
```
dievazqu, line 180: I'm wondering... If a NFT metadata has more fields that these, are we filtering those out?
I think it's not a big deal, because probably we won't use those field in the wallet anyways, but probably we want to return all the field in the metadata.

dievazqu (commented):
Looks good!!
A couple of things before merging:
* Remove the global variable in the nft events
* Wrap the api call in a `try/catch`

My other comments can be ignored.

src/events/NftReceived.ts
MuckT: Fixed in 3ee1fe4ee775beb498b868df8d4cb84fba9c8851!

src/helpers/EventBuilder.ts
MuckT: Fixed in 3ee1fe4ee775beb498b868df8d4cb84fba9c8851!

src/schema.ts
MuckT: I see in the indexer we use
```
export interface NftMetadata extends NftId {
  tokenUri: string | null
  metadata: { [key: string]: any } | null
}
```

I'll have to check how this could be represented in the `schema.ts`

MuckT: It looks like only metadata we can count on always being included are those required by [ERC-721](https://eips.ethereum.org/EIPS/eip-721#specification): `name`, `description` and `image`.

src/helpers/EventBuilder.ts
```diff
@@ -58,19 +61,58 @@ export class EventBuilder {
...
+    const filteredTransfers = transaction.transfers.filter(filterFn)
+
+    // Max number of nfts to fetch at a time for one transaction
+    const concurrencyLimit = 5
+    const promises = filteredTransfers.map(({ tokenId, tokenAddress }) =>
+      EventBuilder.getNft(tokenAddress, tokenId),
+    )
+
+    // Batch promises
+    for (let i = 0; i < promises.length; i += concurrencyLimit) {
```
silasbw, line 100: this batches promise resolution, but not creating the events / work.  https://github.com/valora-inc/blockchain-api/pull/249/files#diff-4e916f1f0fa01c9aa58822c09b649c0a03a30fa292ba1ad19c8f481cdd55d86dR95-R97 creates the events (i.e., pushes new events onto the event queue which could all end up making API requests concurrently).

it's also worth nothing that this approach doesn't do a great job of utilizing the concurrencyLimit because it waits for all of concurrencyLimit promises to resolve before starting another batch of concurrencyLimit promises.  Ideally as soon as 1 promises resolves the logic would create another one so that there's almost always concurrencyLimit running.

anyways, a library can help you address both of these.  I've used p-limit in other repos.  If I could do it over again i'd try using https://github.com/supercharge/promise-pool, probably.

silasbw: p-queue is the same author and maintainer as p-limit. I was worried about the native ESM requirement.

MuckT: I'll give [promise-pool](https://github.com/supercharge/promise-pool) a go.

test/getNft.test.ts
```diff
@@ -0,0 +1,141 @@
...
+})
+
+const mockContractGetter = jest.fn()
+
+jest.mock('../src/utils.ts', () => ({
+  ...(jest.requireActual('../src/utils.ts') as any),
+  getContractAddresses: () => mockContractGetter(),
+}))
+
+nock('https://example.com')
```
kathaypacific, line 19: curious why we need to use nock here instead of just mocking fetch like in other tests?

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +169,33 @@ export class EventBuilder {
   static getWeiForToken(address: string) {
     return Math.pow(10, tokenInfoCache.getDecimalsForToken(address))
   }
+
+  static async getNft(contractAddress: string, tokenId: string): Promise<Nft> {
+    try {
+      const url = `${GET_NFT_API_URL}?contractAddress=${contractAddress}&tokenId=${tokenId}`
+      const response = await fetch(url)
+      if (!response.ok)
+        throw new Error(`Failed to get NFT from ${url} - ${response.status}`)
```
kathaypacific, line 178: nit: this error message combined with the one in the catch block will read `Failed to get NFT for ${contractAddress} - ${tokenId} - Failed to get NFT from ${url} - ${response.status}`...so here we could just have `Received response code ${response.status} from ${GET_NFT_API_URL}`

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +169,33 @@ export class EventBuilder {
...
+      return {
+        tokenId,
+        contractAddress,
+        ownerAddress,
+        tokenUri,
+        metadata,
+      }
+    } catch (error) {
+      logger.error(
+        `Failed to get NFT for ${contractAddress} - ${tokenId} - ${error}`,
```
kathaypacific, line 190: nit: i see most places in this repo pass an object to the logger.error, i think we should do the same here to make it easier to read in the logs. there are a few other places in this PR that we should do the same

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +169,33 @@ export class EventBuilder {
...
+      )
+      return {
```
kathaypacific, line 192: it seems like this function will never throw an error, and instead will return a semi empty object when there is an error....i wonder if it's better to just let this function throw instead, since we are handling the error upstream anyway. i think it's probably sensible to let the consumer decide what to do with errors from this function, rather than always returning this manually created object which makes it harder for a consumer to know if the data is genuine or not

src/helpers/EventBuilder.ts
```diff
@@ -58,19 +62,54 @@ export class EventBuilder {
...
+    const filteredTransfers = transaction.transfers.filter(filterFn)
+
+    const { results, errors } = await PromisePool.withConcurrency(5)
+      .for(filteredTransfers)
+      .process(async ({ tokenAddress, tokenId }) => {
+        return await EventBuilder.getNft(tokenAddress, tokenId)
+      })
+
+    if (errors.length === 0) {
+      nfts.push(...results)
```
kathaypacific, line 100: i'm confused about what we are trying to do here. we are returning possibly more than 1 NFT, but now the client will need to decide which one to display? we only intend to display 1 NFT for the transaction though (according to the designs), so the logic will probably just be "use the first NFT"....so is it worth bothering to return multiple NFTs here? why not just return the first NFT that matches the condition, then we don't need to worry about the batching?

or alternatively, we can build with the assumption that the client will want to display all the NFTs involved in one transaction to give **full** visibility of the transaction in the near future. but then it doesn't make sense to only return NFTs received or sent using the `filterFn` since it paints half the picture, we should instead return all NFTs in the transfer? 

i personally prefer the first option because it is more straightforward for the client and our near term plans. if we go with the second option, we'll need to make it clear in the data whether the nft is coming / going from the wallet.

src/helpers/EventBuilder.ts
```diff
@@ -90,28 +90,25 @@ export class EventBuilder {
 
     const filteredTransfers = transaction.transfers.filter(filterFn)
 
-    // Max number of nfts to fetch at a time for one transaction
-    const concurrencyLimit = 5 
-    const promises = filteredTransfers.map(({ tokenId, tokenAddress }) =>
-      EventBuilder.getNft(tokenId, tokenAddress)
-    )
+    const { results, errors } = await PromisePool.withConcurrency(5)
```
dievazqu, line 93: We already has a library to do [this](https://github.com/valora-inc/blockchain-api/blob/cbc808dc3d8fd06706ad9c7e6b298d3986062268/src/prices/PricesUpdater.ts#L73).
So, probably we should only use 1. We have used tiny-async-pool in other projects as well, so I prefer to use that.

test/blockscoutV2.test.ts
```diff
@@ -189,17 +189,13 @@ describe('NFT events response according to wallet version', () => {
       block: '14203346',
       transactionHash:
         '0xc0b6cb773e2d62c065a3dd96de54c669f377bfb6e2ca51911ea7b27e9523fb0d',
-      tokenId: '1',
-      tokenAddress: '0x6cff0cae0aaec42198d5c59d63d211c0a36a10b6',
```
dievazqu: Why are we removing these fields?

src/helpers/EventBuilder.ts
dievazqu: Yes, I agree with Kathy. Since we don't have designs to support multiple NFTs in the wallet, it's hard to tell how we want to handle the scenario.
For now, I think we should only return 1 NFT, and once we have a better picture of how to handle multiple NFTs, we can come back and add a new field for all the NFTs (or whatever we need).

dievazqu (commented):
Let's keep it simple for now, and only return 1 NFT in the response.

test/blockscoutV2.test.ts
MuckT: These exist on the nft details instead of the transaction level

src/helpers/EventBuilder.ts
MuckT: Moved to [tiny-async-pool](https://www.npmjs.com/package/tiny-async-pool) in https://github.com/valora-inc/blockchain-api/pull/249/commits/7811a19ec369ce186834fbc87b0210ab70a8918d!

src/helpers/EventBuilder.ts
MuckT: There's designs for multiple NFT's [here](https://www.figma.com/file/OxqtolHXtCcuaSOIpTK6gQ/Native-Valora-NFT-experience?type=design&node-id=678-2210&t=rdnvT5CHl63aswNa-0). If we decide not to use them we can always have it wallet just use `nfts[0]`. As this work is already completed, and provides a more complete picture of a transaction I'd rather leave it in.

src/helpers/EventBuilder.ts
MuckT: There are some designs for transactions with multiple nfts: [here](https://www.figma.com/file/OxqtolHXtCcuaSOIpTK6gQ/Native-Valora-NFT-experience?type=design&node-id=678-2210&t=rdnvT5CHl63aswNa-0). Since we're only handling `NFT_RECEIVED` & `NFT_SENT` I don't think we need to worry about the swap nfts case quite yet; I was able to find this [nft swap transaction](https://etherscan.io/tx/0xcf081ab68391fe0be162882ae2931634666c47fc2808d2330e7c674010a176e0) on Ethereum.

test/getNft.test.ts
MuckT: I've spent some time attempting to get `jest-fetch-mock` similar to how it's done in `test/getNft.test.ts`, but it wasn't quite working as expected. Happy to send over a patch.

src/helpers/EventBuilder.ts
kathaypacific: just want to call out that the designs for multiple nfts is not finalised or prioritised, so we shouldn't be designing eng solutions that cater for these designs right now

src/helpers/EventBuilder.ts
kathaypacific: if this service is down, we should throw an error here. if we want to return a semi empty object, the caller of this function should handle the error and return the object. i think this function as it is, is unpredicatible since it swallows the error (depending on the use case, we may want to handle the error differently so it is more flexible to throw the error here)

src/helpers/fetchers.ts
```diff
@@ -0,0 +1,41 @@
+const IPFSGatewayTools = require('@pinata/ipfs-gateway-tools/dist/node')
```
kathaypacific, line 1: why do we need this file? the nft indexer service should handle all of this gateway stuff. why are we duplicating that logic?

src/helpers/EventBuilder.ts
```diff
@@ -130,4 +170,39 @@ export class EventBuilder {
...
+      const { result } = await response.json()
+      const { tokenUri, metadata, ownerAddress } = result
+      return {
+        tokenId,
+        contractAddress,
+        ownerAddress,
+        tokenUri,
+        metadata: {
+          ...metadata,
+          gatewayUrl: handleNftUrl(metadata.image),
```
kathaypacific, line 191: why are we returning the gateway url?

kathaypacific (commented):
hmm i'm a little concerned about the growing complexity of this PR, since yesterday it's grown to also return a gateway url + a bunch of dedicated gateway logic. can we please remove this, and add a follow up PR if needed?

src/helpers/fetchers.ts
MuckT: The Indexer doesn't return the gateway Url, in the NFT viewer the viewer handles it. I tried putting that code in the blockchain API here, but testing the results having the code in the wallet seems to be working better.

I don't think we want this data stored in the DB, as if we changed our gateway settings we'd require a DB migration to adjust the entries.

src/helpers/EventBuilder.ts
MuckT: I moved this code to the indexer in https://github.com/valora-inc/nft-indexer/pull/318.

kathaypacific (approved):
🧑🏻‍🎨 nice!!!

kathaypacific (approved):

PR #255, opened 2023-05-23
feat(refactor): Restructure repo, add type safety, cross-chain compatible datasource interface

jophish:
### Description

For [ACT-788](https://linear.app/valora/issue/ACT-788/create-generic-interface-to-share-between-alchemy-and-blockscout).

This PR does a few inter-related things. The primary goal of this PR was to introduce an interface that can be used by Alchemy as well as Blockscout to create datasources for both. Along the way, I found it necessary to add some rigor to the types we use throughout the repo, as there were a lot of places where types were either not present, or explicitly incorrect. Types have been mostly consolidated into a new `types.ts` file. Additionally, in order to set us up a bit better for multiple different sources of blockchain TX data, I've re-jiggered the directory structure a bit to make things clearer.

The main idea for the interface is that transaction fetching takes 3 main steps:
* Fetching the raw transactions, and shaping them into _some_ internal representation, `T`.
* Classifying the transactions (which requires access to internal fields on `T`)
* Serializing the transactions (which requires access to internal fields on `T`)

While the internal representation `T` is never revealed at the boundaries of the interface, in order to separate the fetching process into logical steps (i.e., functions, whose outputs are piped into the next's input), we need to genericize the class. We specifically _do not_ attempt to have a unified `T` between Blockscout and whatever future datasources we support (namely, Alchemy) since we don't want to attempt to generalize the internal Blockscout transaction representation. We already have a pattern for classification and serialization, but it depends on the particular Blockscout transaction representation we currently use. As such, the `TransactionClassifier` class has also been made generic to match the new interface.

This PR does _not_ update the current `BlockscoutAPI` class to use this new interface, though it should be relatively simple, since `getTokenTransactionsV2` already closely mirrors the 3-step process above. When adding a new datasource, the process will be:
* Figure out a desired TX representation `T`
* Add this TX representation to `TransactionInterface` in `types.ts`
* Add new classifiers/serialization logic under a new directory in `/events`
* Subclass `BaseDataSource`, and implement the required methods
 
### Tested

Unit tested.

### Backwards compatibility

Yes.

### Related issues

- Fixes [ACT-788](https://linear.app/valora/issue/ACT-788/create-generic-interface-to-share-between-alchemy-and-blockscout).


src/events/blockscout/Any.ts
```diff
@@ -0,0 +1,17 @@
+import { metrics } from '../../metrics'
+import { BlockscoutTransaction } from '../../transaction/blockscout/BlockscoutTransaction'
+import { TransactionType } from '../../transaction/TransactionType'
+import { TokenTransactionV2 } from '../../types'
+
+export class Any extends TransactionType<BlockscoutTransaction> {
```
cajubelt, line 6: I'm mid review so still unclear how this is working with the other pieces. but in isolation I'm wondering whether "Unknown" would be a better name for this

src/events/blockscout/EscrowContractCall.ts
```diff
@@ -0,0 +1,23 @@
...
+    return (
+      transaction.transfers.length === 0 &&
+      transaction.input.hasContractCallTo(Contracts.Escrow)
+    )
+  }
+
+  async getEvent(
+    _transaction: BlockscoutTransaction,
+  ): Promise<TokenTransactionV2> {
+    throw new Error('Event type is not serializable')
```
cajubelt, line 17: I'm not seeing the big picture yet but it's a little worrisome to me how often getEvent is throwing an error-- wondering if having it as an abstract method is the right call, given how it only seems to apply to a subset of the events. too early to make a good rec but a tentative idea I'm having is to use composition instead of inheritance here (e.g. TransactionType<BlockscoutTransaction> doesn't require you to implement getEvent, but maybe there is some other interface that does, and some tx types implement that interface as well)

src/events/blockscout/Any.ts
cajubelt: nvm seeing that this was actually just moved

src/events/blockscout/EscrowContractCall.ts
cajubelt: or maybe just returning null sometimes would be better. seems like it used to return undefined some of the time

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
```diff
@@ -0,0 +1,40 @@
+import { ClassifiedTransaction } from '../TransactionClassifier'
+import { BlockscoutTransaction } from './BlockscoutTransaction'
+
+/**
+ * Aggregates Blockscout transactions based on their type, summing up fees, grouping them into one
+ * transaction meaningful for the end user.
+ * Starting with the first transaction adds fees for contract calls to the previous
+ * transaction of another type.
+ */
+export class BlockscoutTransactionAggregator {
```
cajubelt, line 10: I'm not a big fan of static classes. they seem like a workaround for functions not being first class in some languages

cajubelt: or for namespacing. but I don't think either of those reasons are relevant here

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
```diff
@@ -0,0 +1,40 @@
...
+  ): ClassifiedTransaction<BlockscoutTransaction>[] {
+    const aggregatedTransactions = transactions.reduce(
+      BlockscoutTransactionAggregator.aggregateContractCallFees,
+      [],
+    )
+
+    return aggregatedTransactions.filter((t) => t)
+  }
+
+  static aggregateContractCallFees(
```
cajubelt, line 22: is this ever useful elsewhere? or just separate for testing? you could name it with an underscore prefix if it's meant for internal use but still exported for testing

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
```diff
@@ -0,0 +1,40 @@
...
+    accumulator: ClassifiedTransaction<BlockscoutTransaction>[],
+    currentTransaction: ClassifiedTransaction<BlockscoutTransaction>,
+    _currentIndex: number,
+    _array: ClassifiedTransaction<BlockscoutTransaction>[],
+  ): ClassifiedTransaction<BlockscoutTransaction>[] {
```
cajubelt, line 27: why include these if they're not used? a reduce callback with 2 args is very common

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (currentTransaction.type.isAggregatable() && accumulator.length > 0) {
+      const transactionFees = currentTransaction.transaction.fees
+
+      transactionFees.forEach((fee) =>
+        accumulator[accumulator.length - 1].transaction.addFee(fee),
```
cajubelt, line 32: if the previous transaction is not aggregatable, this still modifies it. i lack the understanding to tell whether that's a bug or not

cajubelt: separately, the convention for reduce callbacks is to use pure functions, which means not modifying previous values in the accumulator. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#description paragraph 4.

src/transaction/TransactionAggregator.ts
```diff
@@ -1,39 +0,0 @@
-import { ClassifiedTransaction } from './TransactionClassifier'
-
-/**
- * Aggregates transactions based on their type, summing up fees, grouping them into one
- * transaction meaningful for the end user.
- * Starting with the first transaction adds fees for contract calls to the previous
- * transaction of another type.
- */
-export class TransactionAggregator {
```
cajubelt, line 9: doh, looks like I reviewed old code that was just moved. oh well

src/transaction/TransactionType.ts
```diff
@@ -1,24 +1,27 @@
...
     this.context = context
   }
 
-  abstract matches(transaction: Transaction): boolean
-  abstract getEvent(transaction: Transaction): Promise<any>
-  abstract isAggregatable(): boolean
+  abstract matches(transaction: T): boolean
+  abstract getEvent(transaction: T): Promise<TokenTransactionV2>
+
+  // Since this is a no-op for some transaction types, we default it to false.
```
cajubelt, line 17: dumb question-- what do you mean it's a no-op for some transaction types?

cajubelt: I don't understand why you decided to do this

src/transaction/TransactionType.ts
```diff
@@ -1,24 +1,27 @@
...
+  isAggregatable(): boolean {
+    return false
+  }
 }
 
-export const isTransactionType = (
-  transaction: TransactionType | undefined,
-): transaction is TransactionType => {
+export function isTransactionType<T>(
```
cajubelt, line 23: it looks like this actually does something much more general, namely checks whether something that is K | undefined is actually K and not undefined

src/blockscout.ts
```diff
@@ -235,7 +202,7 @@ export class BlockscoutAPI extends RESTDataSource {
         }),
       )
     )
-      .filter((e) => e)
+      .filter((e): e is TokenTransactionV2 => !!e)
```
cajubelt, line 205: if the types were generalized per my previous suggestion you could re-use isTransactionType here (probly should rename it too)

src/types.ts
```diff
@@ -0,0 +1,241 @@
+import BigNumber from 'bignumber.js'
+import { BlockscoutTransaction } from './transaction/blockscout/BlockscoutTransaction'
+
```
cajubelt, line 3: I'm going to skip this file, lmk if you want me to look at it though

cajubelt (commented):
looks good overall

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
jh2oman: +1 i don't *think* the types should complain if they are removed and they aren't used like charlie said

src/transaction/TransactionType.ts
```diff
@@ -1,24 +1,27 @@
...
-  abstract matches(transaction: Transaction): boolean
-  abstract getEvent(transaction: Transaction): Promise<any>
-  abstract isAggregatable(): boolean
+  abstract matches(transaction: T): boolean
+  abstract getEvent(transaction: T): Promise<TokenTransactionV2>
+
+  // Since this is a no-op for some transaction types, we default it to false.
+  isAggregatable(): boolean {
+    return false
+  }
```
jh2oman, line 20: I think the default to false is just confusing and not obvious when looking at the implementations.

If you do decide to make it back into an abstract, it looks like you could use an abstract property instead
```suggestion
  abstract readonly isAggregatable: boolean
```

jh2oman (commented):
I reviewed it all. Looks really good! not near as daunting as i thought it would be. Charlie asked most of the questions I had. Once those are answered I'll be ready to review again.

src/types.ts
```diff
@@ -0,0 +1,241 @@
...
+  value: BigNumber.Value
+  currencyCode: string
+  exchangeRate: string
+}
+
+export type LegacyEventInterface = LegacyExchangeEvent | LegacyTransferEvent
+
+// A type containing all transaction provider-specific Transaction representations
+// e.g., Blockscout, Alchemy
+export type TransactionInterface = BlockscoutTransaction
```
jophish, line 241: call this "Transaction"

src/blockscout.ts
```diff
@@ -152,7 +119,7 @@ export class BlockscoutAPI extends RESTDataSource {
     address: string,
     afterCursor?: string,
     valoraVersion?: string,
-  ): Promise<TransactionsBatch> {
+  ): Promise<TokenTransactionResult> {
```
jophish, line 122: consider using raw type here or using better name

jophish: `TransactionResult`

src/datasource/BaseDataSource.ts
```diff
@@ -0,0 +1,76 @@
...
+      transactions: serializedTxs,
+      pageInfo,
+    }
+  }
+
+  /**
+   * Fetch raw transaction data from the desired datasource. Returns a list of transactions
+   * in the desired representation for further processing, as well as page information.
+   */
+  abstract fetchRawTxs(
```
jophish, line 53: could technically be private, but need them for tests

src/transaction/TransactionType.ts
jophish: Yeah, aggregation is only required for Blockscout transactions, so I didn't want Alchemy transfer types to be forced to define it. I've subclassed this TransactionType for Blockscout and required that to have an abstract property instead.

src/blockscout.ts
jophish: pretty sure we landed on keeping it as-is

src/transaction/TransactionClassifier.ts
```diff
@@ -1,19 +1,36 @@
...
+import { BlockscoutTransaction } from './blockscout/BlockscoutTransaction'
 
-export interface ClassifiedTransaction<T extends TransactionInterface> {
+export interface ClassifiedTransaction<
+  T extends Transaction,
+  S extends TransactionType<T>,
+> {
   transaction: T
-  type: TransactionType<T>
+  type: S
```
cajubelt, line 11: nit: what about renaming this to transactionType? otherwise it confuses me a little that "type" is not a TS type

cajubelt: it would also be more similar to the TransactionClassifier property named "transactionTypes" that way

src/transaction/blockscout/BlockscoutTransactionAggregator.ts
```diff
@@ -9,8 +8,8 @@ import { BlockscoutTransaction } from './BlockscoutTransaction'
  */
```
cajubelt, line 8:
```suggestion
 * TODO eventually: refactor to not be a static class, which are for namespacing (not needed with modules) or when functions are not first class (not relevant with JS), and either unit test `aggregateContractCallFees` method or combine it with `aggregate`
 */
```

cajubelt: makes sense to punt on these, but it would be nice to record the ideas somewhere so we can pick up where we left off with our thinking on this

cajubelt (approved):
🥳

PR #261, opened 2023-06-01
feat(alchemy): Add skeleton Alchemy datasource 

jophish:
### Description

This is a reboot of my previous PR, which needed to be reverted due to an uncaught issue with a check in the `AlchemyTransaction` constructor.

This updated PR is identical to the previous PR, but with two changes:
* The bug has been fixed, tests have been added, and I've verified that it no longer occurs by spinning up an instance of blockchain-api locally.
* A best-effort implementation of `AlchemyTransaction` was created, rather than leaving it as a TODO.

The `AlchemyTransaction` implementation is important to include now rather than later, since [ACT-791](https://linear.app/valora/issue/ACT-791/transaction-classification-for-alchemy-transfers) and [ACT-792](https://linear.app/valora/issue/ACT-792/serialize-classified-alchemy-transactions) require an actual implementation to work against. The implementation/interface may not be perfect at the moment, but including something now will prevent duplicated work later, and makes the remainder of the implementation tickets parallelizable.

### Tested

Unit tested, and tested locally by spinning up a blockchain-api instance.

### Backwards compatibility

Yes.

### Related issues

Fixes [ACT-790](https://linear.app/valora/issue/ACT-790/create-skeleton-alchemy-data-source).


src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -0,0 +1,46 @@
...
+const ALCHEMY_SUPPORTED_CHAINS = [Chain.Ethereum]
+export class AlchemyDataSource extends BaseDataSource<
+  AlchemyTransaction,
+  TransactionType<AlchemyTransaction>
+> {
+  chain: Chain
+
+  constructor(chain: Chain) {
+    super()
+    if (!ALCHEMY_SUPPORTED_CHAINS.includes(chain)) {
```
jophish, line 21: This was where the previous error was; the check was always failing, causing the datasource instantiation to throw.

src/transaction/alchemy/AlchemyTransaction.ts
```diff
@@ -0,0 +1,89 @@
...
+      throw new Error(
+        'Hashes of all transfers in a single transaction must match the transaction receipt',
+      )
+    }
+  }
+
+  hashesMatch(): boolean {
+    return this.transfersFrom
+      .map((t) => t.hash)
+      .concat(this.transfersTo.map((t) => t.hash))
```
cajubelt, line 62:
```suggestion
      .concat(this.transfersTo)
      .map((t) => t.hash)
```
code golf

cajubelt (approved):

satish-ravi (approved):

src/transaction/alchemy/AlchemyTransaction.ts
jophish: :facepalm: how did i not see this

src/transaction/alchemy/AlchemyTransaction.ts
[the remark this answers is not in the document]
jophish: discussed both of these naming q's offline

PR #264, opened 2023-06-09
feat(alchemy): implement classifyTxs on datasource

satish-ravi:
### Description

Added events for alchemy with matches rule similar to blockscout. Also refactored some fixture data for unit tests

### Tested

Unit tests

### Backwards compatibility

N/A

### Related issues

- Fixes ACT-791


src/datasource/BaseDataSource.ts
```diff
@@ -63,6 +63,7 @@ export abstract class BaseDataSource<
    * Valora version to aid in filtering unsupported transaction types.
    */
   abstract classifyTxs(
+    address: string,
```
satish-ravi, line 66: address is needed to create the context object for the different transaction types

src/events/alchemy/NftReceived.ts
```diff
@@ -0,0 +1,16 @@
+import { TokenTransferV2 } from '../../types'
+import { TransactionType } from '../../transaction/TransactionType'
+import { AlchemyTransaction } from '../../transaction/alchemy/AlchemyTransaction'
+
+export class NftReceived extends TransactionType<AlchemyTransaction> {
+  matches(transaction: AlchemyTransaction): boolean {
+    return (
+      transaction.getNftTransfersTo().length >
```
cajubelt, line 8: this seems like kind of a funny way of classifying nft received. It's like a net inflow of NFT's. But what if you both receive NFT's and send different ones in the same transaction? Isn't that kind of an NFT received too?

cajubelt: I'm not really sure what the implications are of matching like this, so I'm not sure whether I agree with how this is done or not...

test/mock-data/alchemy.ts
```diff
@@ -0,0 +1,95 @@
+import {
+  AssetTransfersWithMetadataResult,
+  AssetTransfersCategory,
+  TransactionReceipt,
+  BigNumber,
+} from 'alchemy-sdk'
+import { AlchemyTransaction } from '../../src/transaction/alchemy/AlchemyTransaction'
+
+export const mockErc20Transfer: AssetTransfersWithMetadataResult = {
```
cajubelt, line 9: would be great to get these from real Alchemy responses to make sure they're realistic

src/events/alchemy/TokenReceived.ts
```diff
@@ -0,0 +1,16 @@
+import { TokenTransferV2 } from '../../types'
+import { TransactionType } from '../../transaction/TransactionType'
+import { AlchemyTransaction } from '../../transaction/alchemy/AlchemyTransaction'
+
+export class TokenReceived extends TransactionType<AlchemyTransaction> {
+  matches(transaction: AlchemyTransaction): boolean {
+    return (
+      transaction.getErc20TransfersTo().length === 1 &&
```
cajubelt, line 8: I guess this makes sense if TokenReceived just represents a simple token transfer where the user is the recipient. Should the NftReceived classifier work the same way though?

cajubelt (commented):
a bit confused by what the different events mean but it looks fine overall

src/events/alchemy/NftReceived.ts
satish-ravi: yeah this in an interesting way of classifying txs for nfts. This is the same logic as what is currently used in the blockscout transaction classifiers. I discussed this with @jophish and we decided to just copy existing logic. Open to other ideas for classifying the txs

cajubelt (approved):

PR #265, opened 2023-06-09
fix(getNfts): add tokenId and filter BlockscoutTokenTransfers

MuckT:
### Description

- Adds `tokenId` to `BlockscoutTokenTransfer`: [Slack thread](https://valora-app.slack.com/archives/C02N3AR2P2S/p1686325941944579).
- Filters BlockscoutTokenTransfer to only those that include a tokenId and contract address before calling `getNft`.
- Should reduce the 504 timeouts seen in [pagerDuty](https://valoraapp.pagerduty.com/incidents/Q1ROK9PERTITQ0).

### Related PR

#255


silasbw (approved):

dievazqu:
Isn't the problem discussed in slack that we are calling `await EventBuilder.getNft(tokenAddress, tokenId)` with `tokenId` as `null`?
How does this solve that?

MuckT:
@dievazqu If the destructing fails to produce a string we will error before we call getNfts. I've tested locally and I am not longer receiving timeouts and I am seeing the expected empty NFT array when there are errors with`tokenId` or `tokenAddress` a.k.a `contractAddress`.

dievazqu (approved):

PR #266, opened 2023-06-13
feat(prices): Adding coingecko

dievazqu:
### Description

Calculating prices from coingecko 

### Tested

Unit tested + locally

### Backwards compatibility

Yes

### Related issues

- Fixes RET-743


src/prices/PricesUpdater.ts
```diff
@@ -107,18 +107,18 @@ export async function storeHistoricalPrices({
...
-  const batchInsertItems = Object.entries(prices)
-    .filter(([token, _]) => token !== cUSDAddress)
-    .map(([token, price]) => ({
-      token,
-      base_token: cUSDAddress,
-      price: price.toString(),
-      at: fetchTime.toISOString(),
-      fetched_from: 'Exchange library',
-    }))
+  // TODO: Fix the base_token and fetch_from values.
```
dievazqu, line 114: This will be addressed in another task

cajubelt: can you add a link to the Linear issue to add context? It's not clear what is broken about the fields just from these comments

src/prices/coingecko.ts
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    const response = await fetch(url)
+    const jsonResponse = await response.json()
+    logger.debug(`Response from coingecko: ${JSON.stringify(jsonResponse)}`)
+    const prices: Record<string, BigNumber> = {}
+
+    for (const tokenAddress of tokensAddresses) {
+      if (jsonResponse[tokenAddress] && jsonResponse[tokenAddress][CURRENCY]) {
+        const price = jsonResponse[tokenAddress][CURRENCY]
+        prices[tokenAddress] = new BigNumber(price)
```
cajubelt, line 24: if we're claiming that the "base token" is cUSD, it seems like we'd need to divide everything by the price of cUSD. otherwise we just add another place in our codebase where we assume cUSD is not depegged.

cajubelt: if you end up doing this, it's probably worth taking the cUSD address as a parameter (like "baseTokenAddress")

src/prices/PricesUpdater.ts
```diff
@@ -156,8 +156,17 @@ async function addStakedCelo(prices: PriceByAddress) {
...
-  return await addStakedCelo(addPeggedPrices(prices))
+  const [exchangeLibraryPrices, coingeckoPrices] = await Promise.all([
+    await exchangeRateManager.calculatecUSDPrices(),
+    await fetchPrices(tokenAddresses),
+  ])
+
+  return await addStakedCelo(
+    addPeggedPrices({
+      ...exchangeLibraryPrices,
+      ...coingeckoPrices,
```
cajubelt, line 169: this seems like it would have a lot of overlap. Is there a plan to make this more efficient at some point? Is there a reason we can't just switch to coingecko now?

src/prices/coingecko.ts
kathaypacific: i agree with this comment but i think there's no concept of base token for coingecko right? fixing the cUSD assumption seems beyond the scope of this PR but would be nice to clarify what our approach will be for that topic - if there's no base token, perhaps storing a null base_token for prices fetched from coingecko would be a good start?

src/prices/PricesUpdater.ts
```diff
@@ -107,18 +107,18 @@ export async function storeHistoricalPrices({
...
-      token,
-      base_token: cUSDAddress,
-      price: price.toString(),
-      at: fetchTime.toISOString(),
-      fetched_from: 'Exchange library',
-    }))
+  // TODO: Fix the base_token and fetch_from values.
+  const batchInsertItems = Object.entries(prices).map(([token, price]) => ({
+    token,
+    base_token: cUSDAddress,
```
cajubelt, line 117: @kathaypacific the base token is hardcoded to cUSD here, which indicated to me that the prices were assumed to be in units of cUSD. the coingecko request only gives us prices in units of usd. So my suggestion was to divide the coingecko prices by coingecko's own price of cUSD, which would then give us prices in units of cUSD and remove the assumption that cUSD is always pegged.

src/prices/coingecko.ts
cajubelt: I replied here to add some context https://github.com/valora-inc/blockchain-api/pull/266/files#r1230301418

cajubelt: A null base_token sounds like it could break dependencies on the firebase node where the price data is stored. But if it doesn't, and the consumers of that data already assume USD if base token is null, then maybe that's what we'd want.

src/prices/coingecko.ts
```diff
@@ -0,0 +1,37 @@
+import BigNumber from 'bignumber.js'
+import { logger } from '../logger'
+import fetch from 'cross-fetch'
+
+export const BASE_URL =
```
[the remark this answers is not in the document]
dievazqu: I think we are fine for now: https://www.coingecko.com/en/api/pricing#:~:text=Our%20Public%20API%20has%20a,the%20next%201%20minute%20window.

src/prices/coingecko.ts
```diff
@@ -0,0 +1,37 @@
...
+  'https://api.coingecko.com/api/v3/simple/token_price/celo'
+
+const CURRENCY = 'usd'
+
+export async function fetchPrices(tokensAddresses: string[]) {
+  const url = `${BASE_URL}?contract_addresses=${tokensAddresses.join(
```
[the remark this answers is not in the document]
dievazqu: That's a good question!
It seems that [250 is the limit](https://support.coingecko.com/hc/en-us/articles/4538868138905-Can-I-get-many-prices-with-one-API-call-#:~:text=Note%3A%20You%20CANNOT%20get%20more,of%20250%20coins%20per%20call.)

Currently we have around 70 tokens, so we are fine, but I'm gonna create a task to batch this query if you are fine with it.

src/prices/coingecko.ts
```diff
@@ -0,0 +1,37 @@
+import BigNumber from 'bignumber.js'
+import { logger } from '../logger'
+import fetch from 'cross-fetch'
+
+export const BASE_URL =
+  'https://api.coingecko.com/api/v3/simple/token_price/celo'
+
+const CURRENCY = 'usd'
+
+export async function fetchPrices(tokensAddresses: string[]) {
```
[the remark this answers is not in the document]
dievazqu: I agree with could extend this function to support multichain, but it's not trivial to know which is the blockchain id in coingecko (we should parse [this](https://api.coingecko.com/api/v3/asset_platforms)).

I can add a `blockchainId` string parameter if you want, but the caller should know which id to pass which I don't like.

I think we can leave it like this, and think how to solve the id mapping in another task since it's not needed for the current one. I think in our models we will have the `chainId`, so we would have to map `chainId` to coingecko id.

test/prices/coingecko.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      json: () => Promise.resolve(mockResponse),
+    })
+    mockFetch.mockReturnValue(mockFetchPromise)
+
+    const prices = await fetchPrices(mockAddresses)
+
+    expect(mockFetch).toHaveBeenCalledWith(
+      `${BASE_URL}?contract_addresses=${mockAddresses.join(
+        ',',
+      )}&vs_currencies=usd`,
```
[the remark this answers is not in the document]
dievazqu: I understand your point and it's true.

But I think it's not better to leave "unmeaningful" constants in the tests, because it's harder for the reviewer/developer to understand if that constant is the expected value or not. So building the expecting value from other values seems a better way to give some meaning to the expected value.

Also, having a constant expected value implies that the tests will break if changing for example the coingecko URL, which might be ok but I think it's more tedious than valuable (but that's subjective I guess).

I understand the idea of reducing the amount of code in the test to reduce bugs, but bugs can be in a malformed constant too. So changing code for constants doesn't necessarily reduce bugs in my opinion.

Also, I think it's harder to maintain large test files that are filled of constants. 

I'll change these tests, but I'm not convinced is an improvement.

src/prices/coingecko.ts
dievazqu: You are right, we should divide everything by the price of cUSD!
My idea was to keep using the `1:1` assumption between `usd` and `cUSD` for now and address that in a [follow up task](https://linear.app/valora/issue/RET-767/[blockchain-api]-based-price-logic-in-usd-instead-of-cusd) (where we break the 1:1 assumption). But it's true that is better to keep the prices in `cUSD` until then 👍

src/prices/PricesUpdater.ts
dievazqu: Yes, the problem is that coingecko doesn't support a lot of the tokens that we are currently supporting.
So our decision was to use coingecko price, and keep using our library for the missing tokens.

I don't think we need to make it more efficient right now, this is running async so I think it's fine.
But if this code becomes a problem there is a lot of room for improvement 👌

src/prices/PricesUpdater.ts
dievazqu: I created [this](https://linear.app/valora/issue/RET-767/[blockchain-api]-based-price-logic-in-usd-instead-of-cusd), but it hasn't have a lot of context as well.

With the comment, I mean that the `base_token` should be `usd` (probably we should rename the field to `base_currency` or a better name?).
But we could keep using the `cUSD` as the base token for all non-cUSD tokens, and have the `usd` price for `cUSD` (we could even use another table for this to avoid mixing concepts in the same table?).

And for the `fetch_from` I was thinking that we should set `Exchange Library` or `Coingecko` depending where the value was calculated from.

dievazqu:
@cajubelt Thanks for reviewing this! And thanks for the comments and suggestions!
I think I have addressed all of them, and the PR is ready for another review

src/prices/coingecko.ts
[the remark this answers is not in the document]
cajubelt: 10 per minute is not very high, do you have any data to support that? for example, looking at historical blockchain-api usage in gcp, are you seeing that we rarely/never exceed 10 queries per minute?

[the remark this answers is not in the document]
cajubelt: from looking around it looks like this part of an endpoint that is called on a cron. do you know how often?

src/prices/PricesUpdater.ts
cajubelt: is fetch_from used programmatically anywhere? analytics, the wallet, ...? seems worth a quick check to make sure an unexpected value won't break anything

src/prices/coingecko.ts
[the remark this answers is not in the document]
cajubelt: it might be worth setting up an alert for errors from coingecko, just to make sure we don't start failing silently here

src/prices/coingecko.ts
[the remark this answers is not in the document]
cajubelt: In the `types.ts` file there is an enum `Chain`. What about adding an object like `ChainToCoingeckoId` and putting it in a constants file? And then adding a `Chain` parameter to this function?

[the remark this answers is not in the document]
cajubelt: you could make this a non-breaking change to the API by taking a chain parameter and defaulting to Celo if it is not provided

src/prices/coingecko.ts
```diff
@@ -0,0 +1,53 @@
...
+    const prices: Record<string, BigNumber> = {}
+
+    const baseTokenUsdPrice = jsonResponse[baseToken]
+      ? new BigNumber(jsonResponse[baseToken][USD])
+      : null
+    if (!baseTokenUsdPrice) {
+      logger.error({
+        type: 'MISSING_BASE_TOKEN_PRICE_FROM_COINGECKO',
+      })
+      return {}
```
cajubelt, line 31: nit: another thing you could do here is just throw an error, since this log-and-return-empty-object logic kinda duplicates your catch block. not a big deal though

cajubelt:
thanks for acting on the feedback!

src/prices/PricesUpdater.ts
dievazqu: Right now it's not used anywhere, I added it to track the source of our prices, but it's not really relevant.

src/prices/coingecko.ts
dievazqu: Yeah, it's true, I thought about adding this to use a different `type`, so it's easier to track.
But I can throw an error with a specific message and handle it in the catch to avoid the duplicated code.

src/prices/coingecko.ts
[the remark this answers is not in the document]
jeanregisser: Were you able to confirm this by calling it locally every minute for maybe 30 mins?
On a side project, I used to call it every minute for a long time (past 2+ years), and it recently stopped working and I had to decrease the frequency a little (or query less tokens).

src/prices/coingecko.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export async function fetchPrices(tokenAddresses: string[], baseToken: string) {
+  try {
+    const url = `${BASE_URL}?contract_addresses=${tokenAddresses.join(
+      ',',
+    )}&vs_currencies=usd`
+    logger.debug(`Fetching prices with ${url}`)
+    const response = await fetch(url)
+    const jsonResponse = await response.json()
+    logger.debug(`Response from coingecko: ${JSON.stringify(jsonResponse)}`)
```
jeanregisser, line 21: Easier to browser in Logs Explorer:
```suggestion
    logger.debug({ jsonResponse }, `Response from coingecko`)
```

src/prices/coingecko.ts
```diff
@@ -0,0 +1,53 @@
...
+import { logger } from '../logger'
+import fetch from 'cross-fetch'
+
+export const BASE_URL =
+  'https://api.coingecko.com/api/v3/simple/token_price/celo'
+
+const USD = 'usd'
+
+/**
+ * Returns token prices compared to the base token.
```
jeanregisser, line 11:
```suggestion
 * Returns token prices in base token units.
```

src/prices/coingecko.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export async function fetchPrices(tokenAddresses: string[], baseToken: string) {
+  try {
+    const url = `${BASE_URL}?contract_addresses=${tokenAddresses.join(
+      ',',
+    )}&vs_currencies=usd`
+    logger.debug(`Fetching prices with ${url}`)
+    const response = await fetch(url)
```
jeanregisser, line 19: Is there a default timeout for this?

src/prices/coingecko.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return prices
+  } catch (e) {
+    const error = e as Error
+    logger.error({
+      error,
+      errorMessage: error.message,
+      type: 'ERROR_FETCHING_PRICES_FROM_COINGECKO',
+    })
```
jeanregisser, line 50: The default [error serializer](https://github.com/trentm/node-bunyan/blob/5c2258ecb1d33ba34bd7fbd6167e33023dc06e40/lib/bunyan.js#L1141) adds more info (stack, name, code) when captured like this:
```suggestion
  } catch (err) {
    logger.error({
      err,
      errorMessage: error.message,
      type: 'ERROR_FETCHING_PRICES_FROM_COINGECKO',
    })
```

jeanregisser (approved):
Awesome! 🚀 🚀 🚀 
Can't wait to see us switching to real USD prices.

Already approved with a couple of small comments / suggestions.

src/prices/coingecko.ts
dievazqu: Good question, it's using the [node-fetch](https://github.com/node-fetch/node-fetch) library, which doesn't have a timeout.
I'll copy the [function](https://github.com/valora-inc/wallet/blob/main/src/utils/fetchWithTimeout.ts) that we have in the wallet to add a timeout here!

dievazqu:
@cajubelt @jeanregisser
I did some changes, mainly:
* I added the chain parameter.
* I moved the try/catch logic outside of the service.
* I added a timeout to the fetching logic.

Just in case you want to take another look.

src/prices/coingecko.ts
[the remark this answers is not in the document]
dievazqu: Good suggestion! I've left it running for like 2 hours now, and everything looks fine.
I think we can release it with the current frequency and lower it in case we start finding problems. Since we have the fallback to the exchange library.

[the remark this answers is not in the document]
dievazqu: Or we can be proactive and lower the frequency now, I think fetching every 5 minutes is reasonable... But I think it's better if we can keep it in 1 minute.

src/helpers/fetchWithTimeout.ts
```diff
@@ -0,0 +1,25 @@
+import fetch from 'cross-fetch'
+
+export const fetchWithTimeout = async (
+  url: string,
+  body: any | null = null,
+  duration: number,
```
cajubelt, line 6:
```suggestion
  duration: number,
  body?: RequestInit
```
using "?" instead of defaulting to null is a nit. But using RequestInit instead of any is not. We should try to keep the benefits of static typing wherever possible instead of using `any`

src/helpers/fetchWithTimeout.ts
```diff
@@ -0,0 +1,25 @@
...
+): Promise<Response> => {
+  const timeout = new Promise<undefined>((resolve, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id)
```
cajubelt, line 10: shouldn't we clear the timeout even if the request succeeds?

cajubelt: also, this leaves the request hanging after timing out, not ideal. There's a SO answer [here](https://stackoverflow.com/a/50101022/5807149) explaining how to improve on this. A similar solution is recommended in the node-fetch docs [here](https://www.npmjs.com/package/node-fetch).

src/prices/PricesUpdater.ts
```diff
@@ -155,13 +156,33 @@ async function addStakedCelo(prices: PriceByAddress) {
   return returnedPrices
 }
 
+async function wrapCoingeckoCall(
```
cajubelt, line 159: I'd probly move this to the coingecko module

cajubelt: also, do you ever want to call `fetchPrices` without wrapping it? not sure it needs its own layer of abstraction

src/prices/coingecko.ts
```diff
@@ -1,53 +1,59 @@
...
-      type: 'ERROR_FETCHING_PRICES_FROM_COINGECKO',
-    })
-    return {}
+  const prices: Record<string, BigNumber> = {}
+  for (const tokenAddress of Object.keys(usdPrices)) {
+    if (usdPrices[tokenAddress]) {
+      prices[tokenAddress] =
+        usdPrices[tokenAddress].dividedBy(baseTokenUsdPrice)
+    }
   }
```
cajubelt:
```suggestion
  for (const [tokenAddress, usdPrice] of Object.entries(usdPrices)) {
    prices[tokenAddress] =
        usdPrice.dividedBy(baseTokenUsdPrice)
  }
```
code golf!

cajubelt (commented):
thanks for adding cross chain compatibility! Looks close. The "any" type and clearing the timeout are the only blocking comments here.

src/prices/PricesUpdater.ts
dievazqu: Sure, I'll move it to the coingecko file.
But yeah, I think the coingecko service should throw an error when it can't fetch prices, and leave the responsibility to handle that error to the consumer. For now, we want to ignore the error, but that might change, so ignoring the error in the service as default didn't sound like the best abstraction.

src/helpers/fetchWithTimeout.ts
dievazqu: Updated! :)

PR #267, opened 2023-06-13
fix(types): sync TokenTransactionV2 types with wallet

cajubelt:
### Description

fees and metadata are required for TokenTransfer type in wallet
    
fees is required for TokenExchange type in wallet

if we continue to leave them as optional here, we run the risk of type errors at runtime when we let the field be undefined for some case where the wallet actually uses them

### Tested

compiled typescript

### Backwards compatibility

yes. Some fields are newly marked required, but no field is newly marked optional. I also verified that there isn't weird logic in the wallet where undefined/null values are treated differently than empty lists for this field. 

### Related issues

na


jophish (commented):
We actually can't make the assumption that `metadata` is always present without further changes. `metadata` is not included on NFT transfer events, as seen [here](https://github.com/valora-inc/blockchain-api/blob/main/src/helpers/EventBuilder.ts#L120). Easy solution is to keep it optional.

jophish (approved):
:facepalm: My bad, didn't see that `metadata` is still optional on NFT transfers, it's just mandatory on token transfers. LGTM!

PR #268, opened 2023-06-13
feat(alchemy): implement getEvent for alchemy TokenReceived

cajubelt:
### Description

implemented getEvent, which will be used for serializing alchemy transactions

### Tested

unit tested

made mock data more realistic using example data from https://docs.alchemy.com/reference/alchemy-getassettransfers#alchemy_getassettransfers

### Backwards compatibility

na

### Related issues


src/events/alchemy/TokenReceived.ts
```diff
@@ -10,7 +11,45 @@ export class TokenReceived extends TransactionType<AlchemyTransaction> {
...
+        blockNum: block,
+        hash: transactionHash,
+        value,
+        rawContract: { address: tokenAddress },
+        to: toAddress,
+        from: fromAddress,
+      },
+    ] = transaction.getErc20TransfersTo()
+    if (!value || !tokenAddress || !toAddress) {
+      throw new Error(
```
cajubelt, line 27: I decided to throw here because i think it would be notably weird and worth logging an error if this took place. we can catch errors in the serialization step where this will be called, and probably just log them. 

An alternative to doing it this way would be to put this check in the match method, and returning false if any required fields are missing, potentially not throwing an error. I'm pretty open to this, interested to hear what reviewers think.

jophish: FWIW I prefer having the check here rather than during matching to get types to work correctly.

src/events/alchemy/TokenReceived.ts
```diff
@@ -10,7 +11,45 @@ export class TokenReceived extends TransactionType<AlchemyTransaction> {
...
+        `getEvent called on TokenReceived with missing required value. transactionHash: ${transactionHash}`,
+      )
+    }
+    const timestamp = new Date(blockTimestamp).getTime()
+    const { name: title, imageUrl: image } =
```
jophish, line 32: Not for this PR I think, but there's a baked-in assumption with this implementation that addresses cross-chain will have the same extra info regardless of chain; I wonder if that's an assumption we can reasonably make, or if our address info will need to be made chain-specific.

jophish (approved):
LGTM!

src/events/alchemy/TokenReceived.ts
cajubelt: @jophish I think the chances of address collisions are negligibly small

PR #269, opened 2023-06-14
feat(alchemy): Add raw Alchemy transaction fetching

jophish:
### Description

For [ACT-789](https://linear.app/valora/issue/ACT-789/implement-raw-tx-fetching-from-alchemy).

Adds raw transaction fetching for the Alchemy datasource. Some background to explain the fetching process I landed on:

Alchemy's `getAssetTransfers` endpoint uses a data model of _asset transfers_, instead of whole blockchain transactions. A single asset transfer consists of the transfer of a single type of asset from one address to another; _blockchain transactions_ consist of any number of such transfers. `getAssetTransfers` can be queried for transfers _from_ or _to_ an address. Since our goal is to aggregate transfers together into _transactions_, the general idea is to call `getAssetTransfers` twice, for transfers from and to the user's address, and aggregate transfers corresponding to identical transaction hashes into a transaction object. Transfers do not contain information about fees (which, after mulling it over for a bit, actually makes decent sense to me), so we augment the aggregated transfers with fee information by fetching transaction receipts via Alchemy's `getTransactionReceipt` RPC API. 

Calling `getAssetTransfers` exactly once for each of transfers to and from a user's address misses out on some edge cases; the results of `getAssetTransfers` are paginated, and it is possible that all the transfers for some transaction are _not_ contained in a single page alone. For the sake of posterity, a "complete" approach where all edge cases are avoided is as follows:
* Get a page of *to* transfers, starting at (and including) the block specified by an afterCursor, if provided
* Get a page of *from* transfers, starting at (and including) the block specified by an afterCursor, if provided
* Find the minimum block number of all transfers among *to* and *from*. This gives us concrete bounds for all transfers we _must_ collect. (`getAssetTransfers` results are ordered from most-to-least-recent)
* Fetch subsequent pages of *to* and *from* pages until we receive transfers with block number less than the min block from above. This guarantees that we've completely specified all transfers for blocks up to and including the min block.
* Merge these transfers, sort by block, and prune based on `MAX_RETURN_RESULTS`
* If pruning actually occurred, set afterCursor to the min block of the pruned results, hasNextPage to `true`

The above is a "complete" approach, but comes at the cost of potentially making many (often unnecessary) requests to Alchemy. Alchemy imposes pretty restrictive limits on API throughput, and `getAssetTransfers` is expensive in terms of throughput. The approach that this PR opts for is as follows:
* Fetch a page of *to* transfers (1000 transfers), starting at (and including) the block specified by an afterCursor, if provided
* Fetch a page of *from* transfers (1000 transfers), starting at (and including) the block specified by an afterCursor, if provided
* Merge these transfers, sort by block, and prune based on `MAX_RETURN_RESULTS`
* If pruning actually occurred, or if Alchemy has more pages available, set afterCursor to the min block of the pruned results, hasNextPage to `true`.

This approach has the advantage of only ever calling Alchemy twice (and is generally simpler) but opens up some _exceedingly_ rare edge cases. If >=1000 transfers for a user occur on a _single block_ (such that one or more of the responses of `getAssetTransfers` contain transfers for just a single block number), some aggregated transactions may be incomplete/missing transfers. 

On a related note, both approaches suffer from a similarly unlikely edge case with regard to pagination. If more than `MAX_RETURN_RESULTS` (currently 25) _transactions_ for a user exist for a _single block_, `afterCursor` will be set to that block. Subsequent fetches using that block as a `startCursor` will fetch transfers beginning with that block (inclusive) and the same results will be returned to the user. This can be avoided in most cases by encoding additional information into the cursor objects.

To summarize, the edge cases we are vulnerable to with this approach include:
* More than 1000 transfers, per-user, per-block
* More than `MAX_RETURN_RESULTS` (25) transactions, per-user, per-block

I really don't see either of these ever realistically happening (esp. for actual, human-being Valora users), which is why I opted for this approach in order to avoid rate limiting from Alchemy. As long as these assumptions hold (that those two edge cases will not occur) subsequent results returned from Alchemy by progressing the `afterCursor` will always be unique, and contain full transactions only. If it's something we're concerned about, it's not difficult to detect when these edge cases are triggered and log a warning in order to track frequency.

Both Alfajores and Mainnet GCP have been updated with secrets containing the Alchemy Ethereum API keys, and these have been added to the deploy yamls.

### Tested

Unit and manual tested by running an instance of blockchain-api locally.

### Backwards compatibility

Yes.

### Related issues

- Fixes [ACT-789](https://linear.app/valora/issue/ACT-789/implement-raw-tx-fetching-from-alchemy).


app.mainnet.yaml
```diff
@@ -13,7 +13,7 @@ env_variables:
   VERIFICATION_REWARDS_ADDRESS: '0x00000000000000000000000000000000000verif'
   WEB3_PROVIDER_URL: 'https://forno.celo.org'
   # Comma separated list of Secrets to load.
-  SECRET_NAMES: 'projects/1027349420744/secrets/blockchain-api/versions/latest,projects/1027349420744/secrets/blockchain-api-db/versions/latest'
+  SECRET_NAMES: 'projects/1027349420744/secrets/blockchain-api/versions/latest,projects/1027349420744/secrets/blockchain-api-db/versions/latest,projects/1027349420744/secrets/alchemy-mainnet-blockchain-api/versions/latest'
```
cajubelt, line 16: currently the secrets are namespaced with the blockchain-api prefix, can you update to match that?

cajubelt: same with alfajores

src/apolloServer.ts
```diff
@@ -21,9 +20,11 @@ export interface DataSources {
 export function initApolloServer({
   currencyConversionAPI,
   pricesService,
+  ethereumDataSource,
```
cajubelt, line 23: I should have mentioned this in a previous PR since we are already doing this in some places. but this does not make it easy to add more chains in the future. Remember Silas's advice of approaching multichain like we're going to keep adding chains and need it to be easy, not as a 1-time thing

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
 
-const ALCHEMY_SUPPORTED_CHAINS = [Chain.Ethereum]
+const ALCHEMY_NETWORK_MAP: Record<string, Record<AlchemyChain, Network>> = {
+  default: {
+    [AlchemyChain.Ethereum]: Network.ETH_GOERLI,
+  },
+  mainnet: {
+    [AlchemyChain.Ethereum]: Network.ETH_MAINNET,
+  },
+  alfajores: {
```
cajubelt, line 37: I find this pretty confusing. Alfajores is a Celo concept. I think the more general word would be "testnet"

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
+    [AlchemyChain.Ethereum]: Network.ETH_GOERLI,
+  },
+}
+
+const TRANSFER_CATEGORIES = [
```
cajubelt, line 42: again, I probably should've said something about this earlier. But this is a relatively long list of transfer types to be including in an MVP. We could conceivably release something that doesn't handle NFT's at all, for instance, and the scope would be lower.

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
 > {
-  chain: Chain
+  chain: AlchemyChain
+  alchemyClient: Alchemy
 
-  constructor(chain: Chain) {
+  constructor({ chain, apiKey }: { chain: Chain; apiKey: string }) {
     super()
-    if (!ALCHEMY_SUPPORTED_CHAINS.includes(chain)) {
+    if (!isAlchemyChain(chain)) {
```
cajubelt, line 71: in general I dislike when a function accepts some type as a parameter, but it always errors out at runtime for certain values. would prefer if this was a static check, like by changing the type of `chain` to `AlchemyChain`

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
       throw new Error(`Unsupported chain for Alchemy: ${chain}`)
     }
     this.chain = chain
+    this.alchemyClient = new Alchemy({
+      apiKey,
+      network:
+        ALCHEMY_NETWORK_MAP?.[DEPLOY_ENV]?.[chain] ??
+        ALCHEMY_NETWORK_MAP['default'][chain],
```
cajubelt, line 79: this default chain behavior seems like a footgun. why not make DEPLOY_ENV required?

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
+
+const TRANSFER_CATEGORIES = [
+  AssetTransfersCategory.ERC721,
+  AssetTransfersCategory.SPECIALNFT,
+  AssetTransfersCategory.ERC20,
+  AssetTransfersCategory.INTERNAL,
+  AssetTransfersCategory.EXTERNAL,
+]
+
+const MAX_ALCHEMY_RESULTS = 1000
```
cajubelt, line 50: why is MAX_ALCHEMY_RESULTS so large when we truncate to 25 transactions? Seems likely that we will see a small number of transfers per transaction, meaning we will probably truncate upwards of 900 transfers every time (at least for users who have that many).

cajubelt (commented):
partial review, skipped around after the top of AlchemyDataSource. there's going to be annoying merge conflicts with my PR

src/datasource/alchemy/AlchemyDataSource.ts
jophish: As an extra safeguard against the edge case I outlined in the PR description; this can realistically be reduced a lot; we can probably take this down to 100~ or so without issue.

test/mock-data/alchemy.ts
```diff
@@ -3,93 +3,338 @@ import {
...
+  ...mockTransferTo,
+  to: mockAddress,
+  category: AssetTransfersCategory.ERC20,
+}
+
+export const mockErc20TransferFrom: AssetTransfersWithMetadataResult = {
+  ...mockTransferTo,
+  from: mockAddress,
+  category: AssetTransfersCategory.ERC20,
+}
```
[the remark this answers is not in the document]
jophish: Whoops, good catch -- these mocks came from cleaning up a merge conflict. Will consolidate.

src/datasource/alchemy/AlchemyDataSource.ts
jophish: Yeah; I can move the type guard outside of the constructor so that we'll have type safety when creating instances.

src/datasource/alchemy/AlchemyDataSource.ts
jophish: Given that we can add the serialization for different event types in a relatively piecemeal fashion, we're still free to do this; we can stop at just token sends, receives, and swaps if we like, though we've already gotten most of the way there.

src/datasource/alchemy/AlchemyDataSource.ts
jophish: Yes, I had the same thought -- I can change the config to use `testnet` and `mainnet` instead, which is more chain agnostic. The GCP naming conventions still remain of course, but that's a bigger issue.

src/apolloServer.ts
jophish: Yeah; ideally this would be a config-only change. I can try to rejigger things to get closer to that goal, and add some instructions to the README for the process of adding a new chain to the repo.

_Really ideally_, we wouldn't have to go through and manually update service-by-service in order to add broad support for a new chain to the entire Valora stack; we could in theory just update one central config source with "new chain info" and all our services would automagically start servicing a new chain. That would require a lot of work, however.

app.mainnet.yaml
jophish: Yep, that's a good callout.

src/datasource/alchemy/AlchemyDataSource.ts
jophish: Agreed; I'll make `DEPLOY_ENV` a required parameter. Probably good to make it an enum value to get type safety as well.

src/apolloServer.ts
cajubelt: yeah that would be cool. it could all just be in Statsig.. haha

cajubelt: for now it would be nice if it could be done in one config-style change per repo

src/datasource/alchemy/AlchemyDataSource.ts
cajubelt: we probably aren't going to be able to rename our GCP projects

src/datasource/alchemy/AlchemyDataSource.ts
cajubelt: we can talk about this tomorrow, but I suspect that NFT serialization is going to be harder

src/datasource/alchemy/AlchemyDataSource.ts
cajubelt: thanks

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -20,26 +27,246 @@ type ClassifiedAlchemyTransaction = ClassifiedTransaction<
...
+          })
+        }
+      }),
+    )
+    return alchemyTransactions.filter(isDefined)
+  }
+
+  /**
+   * Sorts the partial transaction list by date, most-recent to least-recent, and
+   * filters out all but the MAX_RETURN_RESULT most recent entries. Augments transaction
```
cajubelt, line 165:
```suggestion
   * truncates to the MAX_RETURN_RESULT most recent entries. Augments transaction
```

cajubelt (commented):
i've read through everything except the test code now, looking good

src/datasource/alchemy/AlchemyDataSource.ts
```diff
@@ -14,32 +20,262 @@ import {
...
+): chain is AlchemyChain {
+  return Object.values(AlchemyChain).includes(chain as AlchemyChain)
+}
+
+export interface AlchemyAssetTransfers {
+  transfersFrom: AssetTransfersWithMetadataResult[]
+  transfersTo: AssetTransfersWithMetadataResult[]
+}
+
+export class AlchemyDataSourceManager extends RESTDataSource {
```
jophish, line 52: This is now needed as a consequence of a typing update to Apollo, I believe. We cannot return a _map_ of datasources as a datasource in the apollo `dataSources` method in `apolloServer.ts`; every value in the object returned by `dataSources` needs to itself be a data source. Rather than dynamically throwing all the Alchemy datasources in the top level of this object, we have this manager class that acts as an interface to cleanly get Alchemy datasources, which will not pollute the Apollo `dataSources` function as we add more chains.

emulator.env
```diff
@@ -1,4 +1,4 @@
-DEPLOY_ENV=emulator
+DEPLOY_ENV=local
```
cajubelt, line 1: did you mean to push this?

src/index.ts
```diff
@@ -89,30 +90,56 @@ async function parseArgs() {
       type: 'number',
       default: 1.0,
     })
+    .option('deploy-env', {
+      description: 'Deploy environment',
+      type: 'string',
```
cajubelt, line 95:
```suggestion
```
defining "choices" the way you have actually types this as DeployEnv

src/index.ts
```diff
@@ -89,30 +90,56 @@ async function parseArgs() {
...
+      choices: Object.values(DeployEnv),
+      demandOption: true,
+    })
     .epilogue(
       'Always specify arguments as environment variables. Not all arguments are supported as CLI ones yet.',
     ).argv
 
-  return argv
+  const deployEnv = argv['deploy-env'] as DeployEnv
```
cajubelt, line 103:
```suggestion
  const deployEnv = argv['deploy-env']
```
you don't need to type cast if you get rid of the `type: 'string'` part

cajubelt: yargs is smarter than you think!

cajubelt: separately, it looks like there is a mix of yargs function-scoped config parsing and vanilla module-scoped config parsing going on in this repo. from a quick look, it appears that before this change adding deployEnv to the function-scoped config parsing, there at least wasn't any overlap between the config params that were loaded in the 2 different places/ways. 

I find the overlap unsettling.. reminds me of a bug I've seen where the same env var was defined in some unencrypted config file and a secret that was loaded into memory and over-wrote it, except in this case, the bug would be even more devious, because the same config param could be different in the 2 different places it's loaded. (Here, secrets are loaded before reading from process.env, but in config.ts, I think it's the opposite.)

You can punt on this if you want, but I think there should at least be a Linear issue for following up and unifying the way configs are done in blockchain-api. And maybe also a code comment TODO that links to it

cajubelt (approved):

src/index.ts
jophish: Yeah, I was also a bit concerned by the two separate config parsing approaches this repo takes; one locally scoped and one global. I think I'm going to punt for this PR since it's already pretty huge, but will spin up a Linear issue to unify the config approaches.

emulator.env
jophish: Yes, I did. `emulator` as a `DEPLOY_ENV` value I believe has no actual semantics, and now that we're enforcing an enum value for `DEPLOY_ENV`, I wanted to get rid of unnecessary values.

PR #271, opened 2023-06-16
feat(alchemy): TokenSent.getEvent

cajubelt:
### Description

implemented TokenSent.getEvent

NOTE: punts on handling fees, which are paid for in Eth and probably require a breaking change to the API of the service. I am punting on this problem since it was not anticipated as part of this issue and increases scope significantly. See issue [here](https://linear.app/valora/issue/ACT-840/display-fees-correctly-when-paid-in-native-token) 

### Tested

new unit tests

### Backwards compatibility

na

### Related issues


src/events/alchemy/TokenSent.ts
```diff
@@ -10,7 +11,44 @@ export class TokenSent extends TransactionType<AlchemyTransaction> {
...
+      },
+    ] = transaction.getErc20TransfersFrom()
+    if (!value || !tokenAddress || !toAddress) {
+      throw new Error(
+        `getEvent called on TokenSent with missing required value. transactionHash: ${transactionHash}`,
+      )
+    }
+    // get title, image from knownAddressesCache
+    const { name: title, imageUrl: image } =
+      knownAddressesCache.getDisplayInfoFor(toAddress)
```
jh2oman, line 33: So the `address` and `account` are the fromAddress but all the metadata is about the toAddress?

cajubelt: I fixed this

jh2oman (approved):

PR #275, opened 2023-06-20
feat: fetch balances via RPC call

jeanregisser:
### Description

Fetch balances via a RPC (multi)call.

There's a new `FETCH_BALANCES_VIA_BLOCKSCOUT` config variable which controls whether to fetch via blockscout or the new RPC call.

- alfajores uses the RPC call
- mainnet still uses blockscout for now. There will be a follow up PR to enable it there too. But first I want to be sure it behaves correctly on alfajores.

Pros:
- if blockscout is down, the wallet will still be able to fetch balances
- rebasing token balances (mcUSD, etc) will be accurate (blockscout often caches them)
- makes it easy to support other chains

### Tested

**Example before**
```json
{
  "data": {
    "userBalances": {
      "balances": [
        {
          "tokenAddress": "0x048f47d358ec521a6cf384461d674750a3cb58c8",
          "balance": "690000000000",
          "decimals": "10"
        },
        {
          "tokenAddress": "0x10c892a6ec43a53e45d0b916b4b7d383b1b78c0f",
          "balance": "42785365624767809",
          "decimals": "18"
        },
        {
          "tokenAddress": "0x37de72f3a6f53f452d6ef19ecb4e7cf5748572be",
          "balance": "1",
          "decimals": ""
        },
        {
          "tokenAddress": "0x874069fa1eb16d44d622f2e0ca25eea172369bc1",
          "balance": "7711853510182215372",
          "decimals": "18"
        },
        {
          "tokenAddress": "0xbd793c84ddacfd4e48e689cf9dd9cd091348ef08",
          "balance": "1",
          "decimals": ""
        },
        {
          "tokenAddress": "0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9",
          "balance": "35824707859379986302",
          "decimals": "18"
        }
      ]
    }
  }
}
```

**Example after**

```json
{
  "data": {
    "userBalances": {
      "balances": [
        {
          "tokenAddress": "0x048f47d358ec521a6cf384461d674750a3cb58c8",
          "balance": "690000000000",
          "decimals": "10",
        },
        {
          "tokenAddress": "0x10c892a6ec43a53e45d0b916b4b7d383b1b78c0f",
          "balance": "42785365624767809",
          "decimals": "18",
        },
        {
          "tokenAddress": "0x874069fa1eb16d44d622f2e0ca25eea172369bc1",
          "balance": "7711853510182215372",
          "decimals": "18",
        },
        {
          "tokenAddress": "0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9",
          "balance": "35824707859379986302",
          "decimals": "18",
        }
      ]
    }
  }
}
```

### Backwards compatibility

The main difference is that it returns balances only for the known listed tokens.
It is still backward compatible with existing clients, which only care about known tokens anyway.

### A note on performance

Comparing the existing and new implementation, it's quite similar on alfajores:
- existing implementation: around ~500ms querying the dev server, but sometimes when querying quickly it was faster, around 70ms (assuming it was returning cached data)
- new implementation: around ~500ms querying the dev server, when querying quickly ~150ms (assuming it was returning cached data) 

But good surprise, on mainnet it is 5x to 6x faster 🚀 
- existing implementation: around ~600ms querying the dev server, but sometimes into ~900ms
- new implementation: around 100ms querying the dev server

### Related issues

Fixes RET-649


tsconfig.json
```diff
@@ -15,8 +15,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2017",
-    "lib": ["es7", "es2017"]
+    "target": "es2021",
+    "lib": ["es2021"]
```
jeanregisser, line 19: Was needed to be able to use the bigint notation (example: `123n`).

src/blockchain.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    return results
+      .map((result, index) => ({
+        tokenAddress: tokensAddresses[index],
+        balance: result.toString(),
+        decimals: tokensInfo[index].decimals.toString(),
+        symbol: tokensInfo[index].symbol,
+      }))
+      .filter((token) => token.balance !== '0')
```
kathaypacific, line 45: any reason why we don't want to return the balance if it's zero? is it because we are trying to mimic what blockscout does? (was just thinking that the wallet will deal with a zero balance just fine)

kathaypacific (approved):
LGTM!! what is our criteria for turning this on in mainnet? just thinking that it's not great to have alfajores and mainnet being different for too long, if blockscout introduces some changes in alfajores first that don't work then we won't know until it reaches mainnet

src/blockchain.ts
```diff
@@ -0,0 +1,47 @@
...
+    super()
+  }
+
+  async fetchUserBalances(address: string): Promise<UserTokenBalance[]> {
+    const tokensInfo = tokenInfoCache.getTokensInfo()
+    const tokensAddresses = tokensInfo.map((tokenInfo) =>
+      tokenInfo.address.toLowerCase(),
+    )
+
+    const results = await this.client.multicall({
```
kathaypacific, line 25: there's no restriction on how many contracts this multicall can take right? we dont' have to worry about our list of tokens getting too big for one call to handle?

src/blockchain.ts
jeanregisser: Yes, just to save on unnecessary transferred data, otherwise we'd have a long list of 0 balances.

cajubelt (commented):
From a very quick look the implementation looks good, and I look forward to seeing one less dependency on blockscout.

However, we may be jumping from the frying pan to the fire by switching from blockscout to Forno, another cLabs service whose flakiness has affected us recently. Their rate limits brought the indexer to a standstill recently, and it's still unknown to me what those limits are. cLabs' own docs call this out, see "tip" [here](https://docs.celo.org/network/node/forno#what-is-forno), plus a warning about downtime. 

If we more forward with this refactor I'd suggest we do one or more of the following:
- get some more clarity on forno rate limits and uptime guarantees
- get data on how many RPC requests this results in
- (if financially feasible) switch to use QuickNode instead
- fall back to blockscout if the RPC fails (kinda lame but probly better than just failing)

src/blockchain.ts
```diff
@@ -0,0 +1,47 @@
...
+            args: [address as Address],
+          } as const),
+      ),
+      allowFailure: false,
+    })
+
+    return results
+      .map((result, index) => ({
+        tokenAddress: tokensAddresses[index],
+        balance: result.toString(),
```
dievazqu, line 41: I'm a bit confused, I haven't used viem yet... But reading the [documentation](https://viem.sh/docs/contract/multicall.html#return-value), it seems that type of `results` is `({ data: <inferred>, status: 'success' } | { error: string, status: 'reverted' })[]`, but here we are just doing `.toString()` to the object?

dievazqu (approved):
LGTM!
Just a question about the multicall usage.

jeanregisser:
> However, we may be jumping from the frying pan to the fire by switching from blockscout to Forno, another cLabs service whose flakiness has affected us recently. Their rate limits brought the indexer to a standstill recently, and it's still unknown to me what those limits are. cLabs' own docs call this out, see "tip" [here](https://docs.celo.org/network/node/forno#what-is-forno), plus a warning about downtime.
> 
> If we more forward with this refactor I'd suggest we do one or more of the following:
> 
> * get some more clarity on forno rate limits and uptime guarantees
> * get data on how many RPC requests this results in
> * (if financially feasible) switch to use QuickNode instead
> * fall back to blockscout if the RPC fails (kinda lame but probly better than just failing)

Yep very good point, we'll double check these things before enabling this on mainnet.

src/blockchain.ts
jeanregisser: @kathaypacific  I did some more tests related to the [batch size](https://viem.sh/docs/contract/multicall.html#batchsize-optional).

On mainnet (where we have more listed tokens), with the default batch size of 1024 bytes, it sends 6 RPC calls.
Setting it to 0 (no batching), it sends a single RPC call, and forno doesn't complain. 
It's isn't necessarily faster as the single big request can take more time to return, instead of 6 smaller ones.

We can tweak that in a follow up PR when we decide to enable it for mainnet.

PR #277, opened 2023-06-21
chore(legacy-transactions)!: remove deprecated legacy transactions

jh2oman:
### Description
*DO NOT MERGE UNTIL minVersion* is updated in firebase to 1.24

Deprecate legacy transactions! TransactionsV2 have been supported since 1.24 in the app which was released in January 2022. Removing this legacy code will help ease our multichain journey. 

### Tested

*will test with a build of 1.24 to make sure it works*

### Backwards compatibility

Deprecating legacy behavior. not backwards compatible 

### Related issues

https://linear.app/valora/issue/ACT-813/investigateremove-legacy-transaction-fetching


jeanregisser (approved):
Feels good 💥 🚀 💯

PR #280, opened 2023-06-22
fix(currency): Show better Naira exchange rate with bitmama api

jh2oman:
### Description

We have gotten consistent feedback on the ground that our exchange rate for NGN does not reflect reality, so this PR uses bitmama's API specifically for NGN exchange rate calculations.

### Tested

Unit tests and tested running locally, see screenshot

Screen Shot 2023-06-22 at 12 45 48 PM

### Backwards compatibility

n/a

### Related issues

https://linear.app/valora/issue/ACT-820/show-better-naira-exchange-rates


app.alfajores.yaml
```diff
@@ -17,4 +17,5 @@ env_variables:
   STAKED_CELO_TOKEN_ADDRESS: '0xD22E18556E43cb29D6d6172D4b33Fd2Edb629EF2'
   STAKED_CELO_MANAGER_ADDRESS: '0xFfe124dde2b29fA848aD8caAEBE85651F0b5c406'
   GET_NFT_API_URL: 'https://api.alfajores.valora.xyz/getNft'
+  BITMAMA_API: 'https://cico.bitmama.io'
```
cajubelt, line 20: they might have a staging endpoint we can use, though as long as we only use the api for fiat it won't matter

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      const result = await this.get<BitmamaAPIResult>(
+        '/rates/fiat',
+        {
+          from: fromCurrencyCode,
+          to: toCurrencyCode,
+        },
+        {
+          cacheOptions: {
+            ttl: MIN_TTL, // Only actually call the API every MIN_TTL minutes so it doesn't get spammed
```
cajubelt, line 37: minutes? above you said this is in seconds

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+    } catch (error) {
+      logger.error({
+        type: 'ERROR_FETCHING_BITMAMA_EXCHANGE_RATE',
+        error,
+      })
+      // As a fallback, try to return the last cached rate
+      if (this.rateCache.has(`${fromCurrencyCode}-${toCurrencyCode}`)) {
+        return this.rateCache.get(
+          `${fromCurrencyCode}-${toCurrencyCode}`,
+        ) as BigNumber
```
cajubelt, line 55:
```suggestion
      const cachedRate = this.rateCache.get(`${fromCurrencyCode}-${toCurrencyCode}`)
      if (cachedRate) {
        return cachedRate
```
code golf. to avoid the type cast

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+import { logger } from '../logger'
+import { BITMAMA_API } from '../config'
+
+interface BitmamaAPIResult {
+  buy: number
+  sell: number
+}
+
+// ttl in seconds!
+const MIN_TTL = 12 * 3600 // 12 hours
```
cajubelt, line 12: 12 hours seems very conservative. If you Google Naira/USD exchange rates, it's actually more volatile than you'd think-- looks like it's gone up and down by as much as 15% in an hour just in the last week. And we haven't heard anything from Bitmama about rate limits or paid usage of their API. How about 10 minutes instead?

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+export default class BitmamaAPI extends RESTDataSource {
```
cajubelt, line 13:
```suggestion
/**
 * API Client for Bitmama's exchange rate API.
 *
 * Two types of caching are used:
 * - apollo-datasource-rest's built-in expiring cache is used to save a roundtrip and regulate our API usage
 * - a local, non-expiring cache is used as a fallback in case requests fail
 */
export default class BitmamaAPI extends RESTDataSource {
```

cajubelt: took me a bit to understand this, so I figured I'd document it

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+  private rateCache: Map<string, BigNumber>
+  constructor() {
+    super()
+    this.baseURL = BITMAMA_API
+    this.rateCache = new Map<string, BigNumber>()
```
cajubelt, line 18:
```suggestion
    this.rateCache = new Map<string, BigNumber>() // should be ok to not have an eviction policy because we only use this for USD/Naira, and even if we used all 180 currencies in the world, the number of entries has a manageable upper bound
```
I had an argument with myself over whether the potential for a memory leak here was acceptable. I think it's ok, but I suggest we leave this comment here to explain the rationale

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+        `${fromCurrencyCode}-${toCurrencyCode}`,
+        new BigNumber(result.buy),
+      )
+      return new BigNumber(result.buy)
+    } catch (error) {
+      logger.error({
+        type: 'ERROR_FETCHING_BITMAMA_EXCHANGE_RATE',
+        error,
+      })
+      // As a fallback, try to return the last cached rate
```
cajubelt, line 51: it could be interesting to include a timestamp on the cached rate and include that in the error message, so we can see very quickly how far behind we are by using the cached rate

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+          cacheOptions: {
+            ttl: MIN_TTL, // Only actually call the API every MIN_TTL minutes so it doesn't get spammed
+          },
+        },
+      )
+      this.rateCache.set(
+        `${fromCurrencyCode}-${toCurrencyCode}`,
+        new BigNumber(result.buy),
+      )
+      return new BigNumber(result.buy)
```
cajubelt, line 45: if the buy and sell price were identical, then I think this would be fine and we could arbitrarily choose between the two. However, from querying the bitmama API I've seen that they are different. My guess is that the market maker offers 2 different prices to earn money off the spread, and the true market exchange rate is somewhere between the 2 prices. For our use case, we want the market exchange rate. How about using the average of the buy and sell prices instead?

src/currencyConversion/BitmamaAPI.ts
```diff
@@ -1,21 +1,42 @@
...
+      timestamp: number
+    }
+  >
+  constructor(baseUrl: string) {
     super()
-    this.baseURL = BITMAMA_API
-    this.rateCache = new Map<string, BigNumber>()
+    this.baseURL = baseUrl
+    // Should be safe to not have an eviction policy here since we only cache NGN/USD rates
+    // and even eif we used all 200+ currencies, the memory usage would be negligible
```
cajubelt, line 32:
```suggestion
    // and even if we used all combinations of all 200ish currencies, the memory usage would be acceptable
```
idk if I'd call 40k records-- which is the worst case-- negligible

src/index.ts
```diff
@@ -168,7 +176,7 @@ async function main() {
   const exchangeRateAPI = new ExchangeRateAPI({
     exchangeRatesAPIAccessKey: config.exchangeRateApiKey,
   })
-  const bitmamaAPI = new BitmamaAPI()
+  const bitmamaAPI = new BitmamaAPI(args['bitmama-api-base-url'])
```
cajubelt, line 179: note the CI failure on this line. maybe you meant to add it to the output of `parseArgs`?

src/currencyConversion/BitmamaAPI.ts
cajubelt: also I switched from "200+" to "200ish" because the purpose of the comment was to estimate an upper bound. When calculating an upper bound, it doesn't work to use lower bounds (like "at least 200") for the inputs.

src/config.ts
```diff
@@ -56,6 +56,7 @@ export const STAKED_CELO_MANAGER_ADDRESS = (
 ).toLowerCase()
 
 export const GET_NFT_API_URL = process.env.GET_NFT_API_URL
+export const BITMAMA_API = process.env.BITMAMA_API
```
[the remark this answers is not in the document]
cajubelt:
```suggestion
```

cajubelt (approved):

PR #282, opened 2023-06-27
feat(alchemy): seralize alchemy NftReceived

cajubelt:
### Description

- adds Chain to Context, so alchemy transfers know what chain they're on. this allows us to query `getNft` without hardcoding Alchemy nft transfers to Ethereum and Blockscout nft transfers to Celo
- refactors EventBuilder to offer multichain-friendly helper

### Tested

unit tested

### Backwards compatibility

yes

### Related issues


src/helpers/EventBuilder.ts
```diff
@@ -101,34 +111,41 @@ export class EventBuilder {
       .filter(filterFn)
       .filter((transfer) => transfer.tokenAddress && transfer.tokenId)
 
-    const nfts = await asyncPool(
```
cajubelt, line 104: this was moved to getNfts, unit tested, and made more typesafe

src/helpers/EventBuilder.ts
```diff
@@ -19,6 +21,13 @@ import fetch from 'cross-fetch'
 import { GET_NFT_API_URL } from '../config'
 import { logger } from '../logger'
 import asyncPool from 'tiny-async-pool'
+import { isDefined } from '../transaction/TransactionType'
+
+export const ChainToNftNetwork: Record<Chain | AlchemyChain, string> = {
+  // maps Chain to network name used in nft-indexer
+  [Chain.Celo]: 'celo',
+  [Chain.Ethereum]: 'ethereum',
```
satish-ravi, line 29: `Chain.Etherum` is the same as `AlchemyChain.Etherum` because of how we define it in the types, which is why we don't need a separate key for `AlchemyChain.Ethereum` here right? I wonder if we can define `AlchemyChain` as 
```ts
type AlchemyChain = Exclude<Chain, Chain.Celo>
```
and then avoid the `Chain | AlchemyChain` type everywhere ?

src/events/alchemy/NftReceived.ts
```diff
@@ -10,7 +18,29 @@ export class NftReceived extends TransactionType<AlchemyTransaction> {
...
+        .map(({ tokenId, rawContract: { address: tokenAddress } }) => ({
+          tokenId,
+          tokenAddress,
+        }))
+        .filter(hasTokenIdAndTokenAddress),
+      this.context.chain,
+    )
+
+    const timestamp = new Date(
+      nftTransfersTo[0].metadata.blockTimestamp,
```
satish-ravi, line 35: is it safe to assume this value is the same on all `nftTransfersTo`?

src/helpers/EventBuilder.ts
cajubelt: I tinkered with that and decided against it because it's prone to failures if someone ever adds a chain to Chain that is not Alchemy compatible, but forgets to add it to the blacklist in AlchemyChain

cajubelt: personally I was ok with using a union type in a few places to get a bit more bug safety

src/events/alchemy/NftReceived.ts
cajubelt: they are grouped by transaction already at this point

cajubelt: transactions are atomics and belong to the same block

cajubelt: so yes

satish-ravi (approved):

PR #285, opened 2023-06-30
feat(alchemy): NftSent.getEvent

cajubelt:
### Description

serialization function for NftSent

moved shared logic to EventBuilder

refactored types a bit to make sure the new alchemy-specific helper in EventBuilder could require AlchemyChain as the type of its `chain` parameter, which meant Alchemy TransactionType classes needed their `context.chain` field to use AlchemyChain, rather than `Chain | AlchemyChain`

### Tested

unit tests

### Backwards compatibility

na

### Related issues

fixes https://linear.app/valora/issue/ACT-792/serialize-classified-alchemy-transactions


src/events/alchemy/NftReceived.ts
```diff
@@ -20,27 +20,12 @@ export class NftReceived extends TransactionType<AlchemyTransaction> {
...
-          tokenId,
-          tokenAddress,
-        }))
-        .filter(hasTokenIdAndTokenAddress),
-      this.context.chain,
-    )
-
-    const timestamp = new Date(
-      nftTransfersTo[0].metadata.blockTimestamp,
-    ).getTime()
```
cajubelt, line 36: this logic is very similar for NftSent, so I moved it to a shared helper in EventBuilder, mirroring how blockscout events are processed

src/helpers/EventBuilder.ts
```diff
@@ -124,6 +125,48 @@ export class EventBuilder {
...
+
+  static async alchemyNftTransferEvent({
+    nftTransfers,
+    chain,
+    type,
+    transactionHash,
+    block,
+  }: {
+    nftTransfers: AssetTransfersWithMetadataResult[]
+    chain: AlchemyChain
```
cajubelt, line 143: using type AlchemyChain here is what caused most of the other type updates to be required. It wouldn't make sense to accept non-alchemy chains for an alchemy-specific helper, so I refactored the types to make sure Alchemy transactions have Alchemy chains for their `context.chain` member

src/transaction/TransactionType.ts
```diff
@@ -1,15 +1,22 @@
...
 
-export abstract class TransactionType<T extends Transaction> {
-  protected readonly context!: Context
+export abstract class TransactionType<
+  T extends Transaction,
+  C extends Chain | AlchemyChain,
+> {
+  protected readonly context!: {
+    userAddress: string
+    chain: C
```
satish-ravi, line 15: Could we infer the type of chain from T? e.g., `T extends AlchemyTransaction ? AlchemyChain : Chain`
Also separately, maybe it would be cleaner if we have a BlockscoutChain and AlchemyChain instead of Chain and AlchemyChain?

satish-ravi (commented):
Looks good, just had an idea on inferring types

satish-ravi (approved):

PR #288, opened 2023-07-14
feat(datasource): add blockscout data source

satish-ravi:
### Description

Part 1 of moving the blockscout data source to use the new base class. This PR adds a blockscout data source extending from BaseDataSource which is all copied from [getTokenTransactionV2](https://github.com/valora-inc/blockchain-api/blob/72fbe70189ac3f7f1acc830bf2af2928cfcded97/src/blockscout.ts#L121) method in blockscout.ts. Also refactors BaseDataSource to implement serialize txs instead of expecting the subclass to do it since its the same for both blockscout and alchemy

### Tested

CI

### Backwards compatibility

Yes

### Related issues

- Part of ACT-812


test/datasource/BaseDataSource.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      jest
+        .spyOn(datasource, 'fetchRawTxs')
+        .mockResolvedValue({ transactions: mockTxs, pageInfo: mockPageInfo })
+      jest.spyOn(datasource, 'classifyTxs')
+    })
+
+    it('fetches, classifies and returns sorted serialized transactions', async () => {
+      const result = await datasource.getTokenTxs('0xABCD', 'cursor', '1.1')
+
+      // TODO(any) change to 5 once swap serialization is implemented
```
jophish, line 52: Since we're dropping the swap TX currently, can we verify that `logger.warn` is called for it?

jophish (approved):
Assuming that `BlockscoutDataSource.ts` is pretty much entirely copy-paste, LGTM! One suggestion for updating a test.

PR #291, opened 2023-07-27
fix: nft metadata incorrect schema

kathaypacific:
### Description

The `tokenTransactionsV2` query which feeds the valora homefeed returns null metadata for some nft's, even though some metadata is returned by the `/getNft` service for those nft's. e.g. https://valora-app.slack.com/archives/C029Z1QMD7B/p1690384824585089?thread_ts=1689869574.569729&cid=C029Z1QMD7B

Root cause:
The GraphQL schema that we have defined for metadata has the `description` field marked as required. If a metadata object does not contain this field, then that whole object is returned as a null in the response (with an error, in a separate field). It is the fun way that GraphQL works. 

This was affecting the #YourValora nft's, since on the metadata object only the name and imageUrl is defined (but this is what is in the ERC721Metadata spec, so i think it was the schema that was wrong)

### Tested

Locally

### Backwards compatibility

Y

### Related issues

- Fixes RET-793


jeanregisser (approved):
🚀 

Any client type changes needed too?

kathaypacific:
> 🚀
> 
> Any client type changes needed too?

i don't think so - there are no expectations on `description` on the client

